PR #2, opened 2016-09-27
New cv upgrade command

agh1:
Mainly in prep for https://issues.civicrm.org/jira/browse/CRM-19417

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -48,6 +48,7 @@ class Application extends \Symfony\Component\Console\Application {
     $commands[] = new \Civi\Cv\Command\PathCommand();
     $commands[] = new \Civi\Cv\Command\ShowCommand();
     $commands[] = new \Civi\Cv\Command\UpgradeCommand();
+    $commands[] = new \Civi\Cv\Command\UpgradeDbCommand();
     $commands[] = new \Civi\Cv\Command\UpgradeDlCommand();
     $commands[] = new \Civi\Cv\Command\UpgradeGetCommand();
     $commands[] = new \Civi\Cv\Command\UrlCommand();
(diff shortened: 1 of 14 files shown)


totten:
@agh1 I've added a few small fixes. (Described in commit notes.)

One blocker issue in `UpgradeCommand` -- the calls to subprocesses pass through data without proper escaping. I suspect they work in typical usage but would break in some scenarios, like if:
 * The `$problem` report includes any single-quotes.
 * The `$dl['url']` includes any shell characters (eg '&' is common in URLs)
 * The user passes in a `$reportName`  or `$stability` with a shell characters.
 * The `sys_get_temp_dir()` has spaces. (Most Linux systems just use `/tmp` which is fine, but Win/OSX do funny things.)

Suggestion: all variables passed through to `Cv::run()` should be escaped with `escapeshellarg()`.

seanmadsen:
When this is merged, it'd be nice to add an answer to this [SE question](https://civicrm.stackexchange.com/questions/18563/how-do-i-use-cv-to-run-database-upgrades-from-the-command-line) which mentions the new command.

totten:
@agh1 I've made a few updates on this PR

1. *Rebased to match latest `master`* (and force-pushed `agh1:cv-upgrade`)
2. *Significantly revamped the `upgrade:db` command* to support more detailed output. If executed interactively with the default `--out=pretty`, it will display a lot more info. With `-v` or `-vv`, it'll display still more. With `-q`, it'll be quiet. If executed with a structured format (e.g. `--out=json`), it should still return the same content.
3. *Temporarily disabled the other sub-commands.* I think it would be useful to get *something* merged, and we've set the bar rather high by putting all the commands together. I think it'd be an improvement to get this merged with at least one command. (If nothing else, it'll be easier to maintain the code for all the commands... because they'll show up during grep+refactor cycles.)

totten:
I did a local build with #2 and #35 and ran the full `cv` test suite on D7/WordPress/Backdrop, and there don't appear to be any regressions here. 

Let's get this merged already!

totten:
@seanmadsen Posted a comment on https://civicrm.stackexchange.com/a/22949/93 about the `cv upgrade:db` command.

PR #3, opened 2016-11-07
Pop

michaelmcandrew:
Adding support for [pop](https://packagist.org/packages/michaelmcandrew/pop).


tests/Command/PopCommandTest.php
```diff
@@ -0,0 +1,326 @@
+<?php
+
+namespace Civi\Cv\Command;
+use Civi\Cv\Util\Process;
+
+
+class popCommandTest extends \Civi\Cv\CivilTestCase {
```
totten: Small typo ^^

michaelmcandrew:
@totten better now?

michaelmcandrew:
@totten did you spot that with an automated tool i should be using?

totten:
@michaelmcandrew I've been reviewing this during my long train ride home. Some reactions so far:

`1` Some proposed changes in https://github.com/michaelmcandrew/cv/pull/1

`2` The `composer.json` includes this bit which probably doesn't belong?

```
"repositories": [
        {
            "type": "path",
            "url": "/home/michael/src/pop"
        }
    ],
```

`3` When I run the test suite, it fails with an error on `yaml_emit`. That function appear to be defined anywhere in the source tree. Ideas?

```
~/src/cv$ env CV_TEST_BUILD=/Users/totten/buildkit/build/dmaster phpunit4 tests/Command/PopCommandTest.php
PHPUnit 4.8.21 by Sebastian Bergmann and contributors.


Fatal error: Call to undefined function yaml_emit() in /Users/totten/src/cv/tests/Command/PopCommandTest.php on line 32
```

totten:
@michaelmcandrew I bumped into a couple other issues:

 1. The license for `cv` is `MIT`,  but the license for https://github.com/michaelmcandrew/pop is `AGPL-3.0`. That's arguably not changing anything, but I've got a soft preference for 
 2. This is not the fault of `pop`, but an earlier shortcut on my part. `civix` builds on `cv` because it needs to use the same `Bootstrap.php` (and because too many people got confused about installing both executables). This means that... as we add new libraries and features to `cv`, they quietly add deadcode to `civix`. I'm oscillating between whether it's better to (a) combine the programs into one or (b) extract `Bootstrap.php` into an even smaller PHP library.

michaelmcandrew:
@totten 

i merged your pr

i fixed the composer.json.

yaml_emit() is a php function from here: http://php.net/manual/en/function.yaml-emit.php

I decided I prefered it to the symfony yaml component (can't remember why). we could potentially swap it for the symfony component if you think it'll cause people bother (something else to install).

I changed the license to MIT.

not sure what to do about https://github.com/civicrm/cv/pull/3#issuecomment-269530333 part 2

totten:
> did you spot that with an automated tool i should be using?

Yeah -- I use `civilint`, which is basically just [`phpcs`](https://github.com/squizlabs/PHP_CodeSniffer) with the rules from https://github.com/civicrm/coder/

> yaml_emit() is a php function from here: http://php.net/manual/en/function.yaml-emit.php
>
> I decided I prefered it to the symfony yaml component (can't remember why). we could potentially swap it for the symfony component if you think it'll cause people bother (something else to install).

 1. Trying to think what an issue might be with  Symfony YAML -- e.g. maybe there would be a dependency-conflict if you ran `cv pop` on Drupal 8 (which also bundles Symfony YAML)?
 2. The level of bother seems to be platform dependent. If we were only talking about Debian/Ubuntu, then it's probably not much of a bother. In OSX/MAMP, it's a PITA to install C-based PHP extensions -- and I think that's true for many third-party PHP bundles.

> not sure what to do about #3 (comment) part 2

Yeah, I'm really trying to talk myself into either a vision of "fat cv" (mixing most Civi CLI functionality into one) vs "skinny cv" (making it easier to write+distribute multiple CLI tools)...

michaelmcandrew:
> Yeah, I'm really trying to talk myself into either a vision of "fat cv" (mixing most Civi CLI functionality into one) vs "skinny cv" (making it easier to write+distribute multiple CLI tools)...

Can we define some standards for a fat tool that can be extended with multiple skinny tools? is that comment to abstract to be useful?

michaelmcandrew:
@totten. take your point about PITA of C-based PHP extensions. Does this prevent you from using it? Any chance we can merge as it is so people at least get exposed to it and I work on a seperate issue for migrating to symfony's yaml component?

I don't think migrating will be a big deal. I've run in to problems with it on civicrm-docs https://github.com/civicrm/civicrm-docs/issues/8. Wondering about whether to make the switch there (though much smaller audience, so not such a big deal).

totten:
As a general principle, it's OK to ship a copy of `cv` where the `pop` feature only works for some people. We don't need every feature to work in every environment.

As a more specific matter, my daily dev box is one of those environments. :( Feels weird to publish something I can't run...

michaelmcandrew:
The thought had crossed my mind :)

totten:
@michaelmcandrew I needed some fake data today, pulled this up, and did a few tweaks -- listed at the bottom of:

https://github.com/civicrm/cv/compare/master...totten:master-pop?expand=1

I ran into a few issues which didn't seem as simple, though:

1. The `cv.phar` file grows from 2mb to 10mb -- mostly due to the localized Faker data files.
2. When running the `cv.phar pop`, it produces an error (below). Under the theory that the YAML files were missing from `cv.phar`, I updated `box.json` to include them. That didn't seem to be enough -- maybe there's something in the file-loading process that doesn't quite work with PHAR? 

```
Could not create 'Contact' [Mandatory key(s) missing from params array: one of (first_name, last_name, email, display_name)] (fields below)
Array
(
    [contact_type] => Individual
)
```

totten:
I think we can get this mergeable with two more changes:
 1. Update `box.json` to only include the default locale (eg `vendor/fzaninotto/faker/src/Faker/Provider/en_US/`).
 2. Update `michaelmcandrew/pop` to support loading data files from the PHAR. (Guess: there doesn't seem to be a lot of file I/O in `Civi\Pop\Pop` besides `yaml_parse_file(...)`. Maybe change that to `yaml_parse(file_get_contents(...))`?)

Although if you'd rather have all the locales, then we should talk about making `pop` a standalone tool. (I'd help with some refactoring in `cv` and Jenkins setup, but we'd want to talk about the approach.)

totten:
Lingered too long. Even if we did both of my suggestions above, it would still be suboptimal -- because it would be missing datasets for other locales. (And supporting other locales is a good thing...)

Closing and suggesting an alternative (https://github.com/michaelmcandrew/pop/pull/7) which should be more maintainable and allow better l10n data.

PR #5, opened 2016-12-25
instruct on installing without sudo

ErichBSchulz:
not sure if non-sudo installation will have other implications... but nice to avoid sudo?


totten:
Agree that it's nice to avoid `sudo`, but see https://github.com/totten/civix/pull/85/files#r93977215 for (a) rationale behind this approach and (b) a suggested alternative way to address this problem (i.e. publish+hyperlink a general guide for PHAR installation).

PR #6, opened 2016-12-31
Feature/verbose bootstrap

ErichBSchulz:
Hopefully this is a step in the right direction.

I wrote up other aspects of dealing with silent failure here:

http://civicrm.stackexchange.com/questions/9091/civix-civicrmping-not-printing-any-output/


totten:
@ErichBSchulz Merged some additional changes in #7

PR #14, opened 2017-01-19
Implement ExtensionPathCommand

totten:
This would be seem to be useful for writing simpler developer docs.
Instead of asking the user to go look up a path, you'd give them
a command to do it, e.g.

```
cd `cv ext:path`
civix generate:module org.example.foo
```

or

```
cd `cv ext:path flexmailer`
civix generate:upgrader
```

However, I'm unresolved on whether this is the best formulation for those
commands.  It feels like something a bit more generic might be better, but
I'm having trouble getting a formulation that feels perfect+simple.

Some ideas

```
cd `cv path extensionsDir`
cd `cv path imageUploadDir`
cd `cv path '[civicrm.files]'`
cd `cv path '[civicrm.root]'`
cd `cv path '[cms.root]'`

cd `cv ext:info flexmailer --path`

echo "Please send email to: "
cv ext:info flexmailer -x maintainer/author
```

The `ext:list` is already pretty generic and may actually give the desired
info, although it a bit of typing:

```
cd `cv ext:list -L /^flexmailer/ --columns=path --out=list`
```

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -37,6 +37,7 @@ class Application extends \Symfony\Component\Console\Application {
     $commands[] = new \Civi\Cv\Command\ExtensionEnableCommand();
     $commands[] = new \Civi\Cv\Command\ExtensionDisableCommand();
     $commands[] = new \Civi\Cv\Command\ExtensionListCommand();
+    $commands[] = new \Civi\Cv\Command\ExtensionPathCommand();
     $commands[] = new \Civi\Cv\Command\ExtensionUninstallCommand();
     $commands[] = new \Civi\Cv\Command\ExtensionUpgradeDbCommand();
     $commands[] = new \Civi\Cv\Command\FillCommand();
--- a/src/Command/ExtensionPathCommand.php
+++ b/src/Command/ExtensionPathCommand.php
@@ -1,0 +1,70 @@
+<?php
+namespace Civi\Cv\Command;
+
+use Civi\Cv\Application;
+use Civi\Cv\Encoder;
+use Civi\Cv\Util\ExtensionUtil;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+
+class ExtensionPathCommand extends BaseExtensionCommand {
+
+  /**
+   * @param string|null $name
+   */
+  public function __construct($name = NULL) {
+    parent::__construct($name);
+  }
+
+  protected function configure() {
+    $this
+      ->setName('ext:path')
+      ->setAliases(array())
+      ->setDescription('Look up an extension path')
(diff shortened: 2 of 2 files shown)


totten:
OK, I think I like this formulation -- implement one `cv path` command with options for looking up paths from different sources, i.e.

 * Extension paths (`-x`; basically, `CRM_Core_Resources::singleton()->getPath($key)`)
 * Dynamic path expressions (`-d`; basically, `Civi::paths()->getPath($key)`)
 * Settings (`-s`; basically, `Civi::paths()->getPath(Civi::settings()->get($key))`)

As in:

```
$ cv path -x=flexmailer
/var/www/sites/default/files/civicrm/ext/org.civicrm.flexmailer
$ cv path -x=org.civicrm.flexmailer
/var/www/sites/default/files/civicrm/ext/org.civicrm.flexmailer
$ cv path -s=extensionsDir
/var/www/sites/default/files/civicrm/ext
$ cv path -s=templateCompileDir
/var/www/sites/default/files/civicrm/templates_c
$ cv path -d='[civicrm.root]/packages'
/var/www/sites/all/modules/civicrm/packages
$ cv path -d='[cms.root]/misc'
/var/www/misc
$ cv path
/var/www
$ cv path --all
+-----------+------------------------------+-----------------------------------------------------------------+
| type      | name                         | path                                                            |
+-----------+------------------------------+-----------------------------------------------------------------+
| dynamic   | [civicrm.files]              | /var/www/sites/default/files/civicrm                            |
| dynamic   | [civicrm.root]               | /var/www/sites/all/modules/civicrm                              |
| dynamic   | [cms.root]                   | /var/www/sites                                                  |
| extension | org.civicrm.bootstrapcivicrm | /var/www/sites/vendor/civihr/bootstrapcivicrm                   |
| extension | org.civicrm.flexmailer       | /var/www/sites/default/files/civicrm/ext/org.civicrm.flexmailer |
| setting   | extensionsDir                | /var/www/sites/default/files/civicrm/ext                        |
| setting   | imageUploadDir               | /var/www/sites/default/files/civicrm/persist/contribute         |
| setting   | templateCompileDir           | /var/www/sites/default/files/civicrm/templates_c                |
+-----------+------------------------------+-----------------------------------------------------------------+
|
```

Notes:
 * When using this in a script, you should probably choose the most precise formulation you can (e.g. prefer `-x=...` instead of `-s=extensionsDir`; prefer `-s=extensionsDir` over `-d=[civicrm.files]/ext`).
 * We can have parallel options in `cv path` and `cv url` because the options (`-x`/`-s`/`-d`)
   don't conflict with the current semantics of `cv url`. The main constraint from `cv url` is that
   the default output should be the CMS root.
 * The output mode usually defaults to `list`, unless you specify `--all`.
 * For `-d` on v4.6, it can't work.
 * For `-s` on v4.6, probably use `CRM_Core_Config` instead.
 * No, this doesn't sound simple, but that's because paths in civicrm-core aren't simple. It's hopefully more transparent, though.

ErichBSchulz:
I like this! I was having similar thinks!

PR #16, opened 2017-02-14
Implement PathCommand

totten:
This would be useful for writing simpler developer docs. Instead of asking the user to go look up a path, you'd give them a command to do it, e.g.

```
cd `cv path -x.`
civix generate:module org.example.foo
```

or

```
cd `cv path -x flexmailer`
civix generate:upgrader
```
or

```
cd `cv path -c templateCompileDir`
```

Generally, this has a few main properties:
 * In CLI usage, the default output is `list` which should be suitable for bash scripting.
 * You can manually set output using `--out` and `--format` (similar to `ext:path` command).
 * If you call `cv path` from a PHP or Javascript wrapper, it should use an env var to change the default output (e.g. `json`).
 * There's more than one way to look up a path:
   * Extension paths (`cv path -x cividiscount`; roughly like `CRM_Core_Resources::singleton()->getPath($key)`)
   * Config properties (`cv path -c templateCompileDir`; roughly, `CRM_Core_Config::singleton()->{$key}`)
   * Dynamic path expressions (`cv path -d '[cms.root]'`; roughly, `Civi::paths()->getPath($key)`)
 * The command is generally for looking up folders. If you need a specific file, then you can also add the files to the list.

(This is a replacement for PR #14.)

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -37,6 +37,7 @@ class Application extends \Symfony\Component\Console\Application {
     $commands[] = new \Civi\Cv\Command\ExtensionEnableCommand();
     $commands[] = new \Civi\Cv\Command\ExtensionDisableCommand();
     $commands[] = new \Civi\Cv\Command\ExtensionListCommand();
+    $commands[] = new \Civi\Cv\Command\ExtensionPathCommand();
     $commands[] = new \Civi\Cv\Command\ExtensionUninstallCommand();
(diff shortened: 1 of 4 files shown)


totten:
On the notation for files, it'd probably be more intuitive to change from `cv path -x cividiscount info.xml` to `cv path -x cividiscount/info.xml`.

PR #21, opened 2017-05-11
CRM-18799 remove use of exec in cv

kcristiano:


totten:
This appears to be a port of https://github.com/civicrm/civicrm-wordpress/pull/108, which was previously merged into `civicrm-wordpress`. That code has since been in a number of releases, and no one seems to have changed it since then, so that's positive.

OTOH, I did a quick spot-check on my local `dmaster` using a basic test command (`cv ev 'return CRM_Utils_System::version();'`). Without the patch, it works:

```
~/buildkit/build/dmaster/sites$ ~/src/cv/bin/cv ev 'return CRM_Utils_System::version();'
"4.7.31"
```

But with the patch, it fails:

```
~/buildkit/build/dmaster/sites$ ~/src/cv/bin/cv ev 'return CRM_Utils_System::version();'

  [Exception]
  Failed to locate civicrm.settings.php. By default, this tool searches the parent directories for a standard CMS (Drupal, WordPress, etal) and standard civi
  crm.settings.php. Symlinks and multisite configurations may interfere. To customize, set variable CIVICRM_SETTINGS to point to the preferred civicrm.settin
  gs.php.


php:eval [--out OUT] [--level LEVEL] [-t|--test] [-U|--user USER] [--] [<code>]
```

kcristiano:
@totten Thanks for the feedback.  It was a straight port to be consistent between wp-cli and cv.   My current thinking is that we don't need to exclude exec from cv.  If cv is used I would expect that the command is available.

If we do want to be consistent I can take another look at this.

PR #26, opened 2017-12-08
fixes #25 allow install with symfony 3

### LINKED ISSUE #25: does cv work with symfony 3.x
Wondering if cv really needs symfony 2.8 or can work w/ symfony 3.x?

(actually even symfony 4.x is out now...:)

mfb:
I didn't find any problems running cv with Symfony 3 (but it would be a good idea to run tests to verify this)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,10 +10,10 @@
     ],
     "require": {
         "php": ">=5.3.3",
-        "symfony/console": "~2.8",
-        "symfony/process": "~2.8",
-        "symfony/filesystem": "~2.8",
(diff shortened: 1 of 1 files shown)


mfb:
@totten let me know if you have any thoughts on this one

totten:
Thanks, @mfb. I've run it through the test-suite and confirmed it still works. :+1: 

Notably, the main distribution files (`phar`) would be committed to a particular version, and they'll follow whatever's in the `composer.lock`. But for anyone who downloads directly via `composer`, it's good to loosen up the dependency declaration.

PR #31, opened 2018-01-11
Implement `core:install` and `core:uninstall` commands

totten:
__Context__: I'm working on installer customizations for the would-be build
intended for wordpress.org.  But I need ways to quickly (re)try the new
installer library, and this helps.

See also: https://github.com/civicrm/civicrm-setup/


totten:
For sysadmins, there's CLI help (eg `cv core:install -h`).

For developers interested in the nitty-gritty, see https://github.com/civicrm/civicrm-setup/blob/master/docs/getting-started.md

PR #32, opened 2018-01-22
Require PHP v5.4+

totten:
The recent #31 introduced the use of traits, which will fail to load on PHP 5.3 or earlier. Moreover,  `civicrm-core` now requires PHP 5.4+. 

When running on an older system, we should fail with pleasant error message.

Just in case anyone *hasn't* transitioned yet, I've also added a link in the README to the last version that supports PHP 5.3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ Requirements
 
 A local CiviCRM installation.
 
+PHP v5.4+.
+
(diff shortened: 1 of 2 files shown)


colemanw:
Well, the "pleasant message" is going to be "Parse error: invalid syntax" due to the use of short array syntax. But next month when we require 5.5 it can be more pleasant :)

PR #35, opened 2018-02-16
Allow booting any command with CMS-first protocols

totten:
Currently, most commands extend `BootTrait` (Civi-first boot), and a couple
commands extend `CmsBootTrait` (CMS-first boot).

This patch combines `BootTrait` and `CmsBootTrait` into one utility, thus
making the various boot mechanisms more interchangeable. For example, you
can now run an arbitrary snippet using any of these:

 * `cv ev 'echo "Hello\n";' --level=classloader`
 * `cv ev 'echo "Hello\n";' --level=full`
 * `cv ev 'echo "Hello\n";' --level=cms-only`
 * `cv ev 'echo "Hello\n";' --level=cms-full`

As a matter of policy, the default boot-levels for various commands remain
unchanged.

(This is mostly written to help me test different boot styles on D8 for https://lab.civicrm.org/dev/drupal/issues/5 , but it's also a small increment toward https://lab.civicrm.org/dev/cloud-native/issues/7.)


totten:
I did a local build with #2 and #35 and ran the full cv test suite on D7/WordPress/Backdrop, and there don't appear to be any obvious regressions.

PR #38, opened 2018-03-10
Add a --cwd option to set current directory

### LINKED ISSUE #37: add a --cwd option to set current directory
It would be easier to use cv in automated scripts if we added a --cwd option, which various utilities such as composer, drush, npm, etc. have (in composer this is --working-dir not --cwd)

mfb:
fixes #37

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -24,6 +24,31 @@ class Application extends \Symfony\Component\Console\Application {
   }
 
   /**
+   * {@inheritdoc}
+   */
+  protected function getDefaultInputDefinition() {
+    $definition = parent::getDefaultInputDefinition();
+    $definition->addOption(new InputOption('cwd', NULL, InputOption::VALUE_REQUIRED, 'If specified, use the given directory as working directory.'));
+    return $definition;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function doRun(InputInterface $input, OutputInterface $output) {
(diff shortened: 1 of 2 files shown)


totten:
Cool. I think this is amenable to adding some kind of test-case -- perhaps in [EvalCommandTest](https://github.com/civicrm/cv/blob/master/tests/Command/EvalCommandTest.php), add a scenario like this pseudocode?

```php
  public function testEvalWithCwdOption() {
    // go somewhere else... somewhere that cv won't work...
    chdir(sys_get_temp_dir()); 

    $cwdOpt = "--cwd=" . escapeshellarg($this->getExampleDir()); 
    $helloPhp = escapeshellarg('printf("eval says version is %s\n", CRM_Utils_System::version());');
    $p = Process::runOk($this->cv("ev $cwdOpt $helloPhp"));
    $this->assertRegExp('/^eval says version is [0-9a-z\.]+\s*$/', $p->getOutput());
  }
```

mfb:
I added a test... was that pseudocode or working code?

totten:
The test suite passes, and I was also able to use the feature locally. :)

>  was that pseudocode or working code?

Ooops, well, guess it turned out to be working code. 😜 I was expecting that something off-the-cuff would fail and need twiddling, but sometimes I guess lighting strikes your pseudocode.

Thanks for updating it and pinging again!

PR #43, opened 2018-06-11
closes #42 - fixes ConsoleQueueRunner

### LINKED ISSUE #42: cv upgrade:db fails
I've seen this error on the last several releases, both WP and Drupal, when running `cv upgrade:db`:

```
  [Symfony\Component\Debug\Exception\FatalThrowableError]                                                                                                                                                          
  Type error: Argument 1 passed to CRM_Upgrade_Form::doFileCleanup() must be an instance of CRM_Queue_TaskContext, null given, called in /home/jon/local/crm/htdocs/sites/all/modules/civicrm/CRM/Queue/Task.php   
   on line 88  
```

MegaphoneJon:
This is admittedly a late-night hack - but it seems like a straightforward fix, and I can confirm it fixes the issue in #42.

### DIFF
--- a/src/Util/ConsoleQueueRunner.php
+++ b/src/Util/ConsoleQueueRunner.php
@@ -73,7 +73,7 @@ class ConsoleQueueRunner {
 
       if (!$this->dryRun) {
         try {
-          $task->run();
+          $task->run($taskCtx);
         }
         catch (\Exception $e) {
           // WISHLIST: For interactive mode, perhaps allow retry/skip?


MegaphoneJon:
@totten Is there anything I can do to help speed along review?  This should be straightforward, because the statement as written doesn't conform to the signature of the function it's calling.

totten:
Ah, thanks for the ping @MegaphoneJon. Agree that this improves compliance with the contract. Merged and tagged as v0.2.5. New build of `cv.phar` has been published.

PR #44, opened 2018-10-31
quick and dirty cv core:download

michaelmcandrew:
I wasn't sure why there is no core:download command for cv (hopefully, I didn't miss something).

wp-cli and drush have equivilant commands so I wrote this quick and dirty version for cv.

Initial thoughts? Happy to neaten up, refactor, etc.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -85,6 +85,7 @@ class Application extends \Symfony\Component\Console\Application {
       $commands[] = new \Civi\Cv\Command\ScriptCommand();
       $commands[] = new \Civi\Cv\Command\CoreCheckReqCommand();
       $commands[] = new \Civi\Cv\Command\CoreInstallCommand();
+      $commands[] = new \Civi\Cv\Command\CoreDownloadCommand();
       $commands[] = new \Civi\Cv\Command\CoreUninstallCommand();
     }
     return $commands;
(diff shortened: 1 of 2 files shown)


michaelmcandrew:
Thanks for your thoughts.

OK - so it has been thought about to some extent already - just within a different context, which i guess is why I didn't find it.

> I wonder if it's better to have core:download and upgrade:dl as separate commands -- or to combine them?

From a usability perspective, for people that are looking to create initial set up workflows, having a command called `core:download` (even if it shares underlying implementation with `upgrade:dl`) would be good IMO

FWIW I used `release` rather than `ver`/`version` since version is kind of reserved in cli tools.

The stories look good. We probably want to make downloading l10n optional (and defaulted to on?)

Seems like upgrade is some combination of this command with a "force" to overwrite and some form of "auto detect location of existing CiviCRM download".

Seems like abstracting the download stage is also good. I didn't look at how the extensions downloading (and caching?) operates but maybe some refactoring of that would be good.

Story 3 looks like the easiest one to start with.

michaelmcandrew:
PS. the context is that I am creating Dockerfiles for CiviCRM on WordPress and Drupal and I want commands to insert round about here:

https://github.com/michaelmcandrew/civicrm-docker/blob/master/wordpress/Dockerfile#L10
https://github.com/michaelmcandrew/civicrm-docker/blob/master/drupal/Dockerfile#L11

i.e. a non interactive scripted environment with uninstalled CMSes, which is the reason that I would be interested in starting with 3.

> Story 3 looks like the easiest one to start with.

I probably should have said, looks like the most useful story for me :)

michaelmcandrew:
I would still quite like it :)

PR #45, opened 2018-12-04
CRM-19115 - PHP and MySQL timezones should match

totten:
The general idea is to resync the MySQL timezone as soon as we've reached a
fairly high level of bootstrap (i.e.  once Civi+CMS are both loaded and the
current user is set).

* For the CMS-first protocol, we can put the fix directly into `CmsBootstrap.php`
* For the Civi-first protocol, the `Bootstrap.php` stops a bit sooner (i.e. 
  before the CMS or user has been loaded).

This fixes the problem where `testConcurrency` fails in some environments.
The test launches `cv api job.process_mailing`; but if the timezones aren't
sync'd correctly, then the filter on `civicrm_mailing_job.scheduled_date`
misbehaves.

Related: https://issues.civicrm.org/jira/browse/CRM-19115

### DIFF
--- a/src/CmsBootstrap.php
+++ b/src/CmsBootstrap.php
@@ -183,6 +183,11 @@ class CmsBootstrap {
       $this->ensureUserContact();
     }
 
+    // Some UF integrations/versions don't seem to do this... work-around...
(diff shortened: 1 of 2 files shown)


seamuslee001:
@totten i have tested this on my system and confirmed it works correctly approved

PR #47, opened 2019-06-05
support D8 in the project root as well as webroot

MegaphoneJon:
D8 has a "project root", which is the folder above the "webroot", to support the `vendor` directory outside the webroot, etc.  Drush 9+ can detect your install anywhere inside the project root, not just the webroot, and it'd be nice for `cv` to do the same.


totten:
Cheers @MegaphoneJon ! I've confirmed that this does work in the project root as well as the web root.

totten:
Aside: I extended/generalized a bit further with af44d6f7a897a0696125dd5742566d5962d1714e so that (a) it works with `core:install` (and other CMS-first boots) (b) it handles the `web/` structure similarly in other CMSs.

PR #50, opened 2019-09-16
cv: print a more helpful error message

lynxlynxlynx:
I'm still debugging why it works in my shell, but not through cron and since it breaks here,
it'd be useful to know what it thinks the interface is.


mattwire:
@totten please can we merge this?

totten:
Merged. New build of `cv.phar` published (`v0.2.22`) and assimilated into current buildkit.

PR #52, opened 2019-09-26
Don't assume _ENV is available in example boot script, prefer getenv()

artfulrobot:
The example boot script assumes that the superglobal `$_ENV` is populated. However PHP now recommends that this variable is NOT normally included in [`variables_order`](https://www.php.net/manual/en/ini.core.php#ini.variables-order) configuration.

> There is a performance penalty paid for the registration of these arrays and because ENV is not as commonly used as the others, ENV is not recommended on productions servers. You can still get access to the environment variables through getenv() should you need to.
>
> Source: `/etc/php/7.2/cli/php.ini` (as packaged by Debian)

By switching the example code to use `getenv()` it is more robust and does not depend on a particular configuration of the server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ and output parsing:
 function cv($cmd, $decode = 'json') {
   $cmd = 'cv ' . $cmd;
   $descriptorSpec = array(0 => array("pipe", "r"), 1 => array("pipe", "w"), 2 => STDERR);
-  $env = getenv() + array('CV_OUTPUT' => 'json');
+  $env = (function_exists('getenv') ? getenv() : $_ENV) + array('CV_OUTPUT' => 'json');
   $process = proc_open($cmd, $descriptorSpec, $pipes, __DIR__, $env);
   fclose($pipes[0]);
   $result = stream_get_contents($pipes[1]);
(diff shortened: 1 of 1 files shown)


totten:
Oh, very handy that `getenv()` now returns the full environment. Looks like that was added in PHP 7.1 (https://www.php.net/getenv).

Civi still has releases supporting PHP 5.6 and PHP 7.0.  When I ran `queuetest` before, I had to patch this out so that I could run on `php70` (which is the compatibility target for new versions of Civi). Could you update it with a fallback behavior -- i.e. "use `$_ENV` or `getenv()`, whichever is available"?

totten:
I tested 08354c8 on php70 and php72, and it was problematic on php70. NB `getenv()` does exist in PHP 4+, but the older versions require the `$varname` param. If we flip the test the other way (36e6010), then it works.

* PHP 7.2
    * `php -d variables_order=EGPS testscript.php` => Works.
    * `php -d variables_order=GPS testscript.php` => Works. Improvement.
* PHP 7.0
    * `php -d variables_order=EGPS testscript.php` => Works
    * `php -d variables_order=GPS testscript.php` => Fails. The symptom is a little different from before, but the outcome is the same as ever.

PR #57, opened 2019-12-18
Changes for drupal7 multisite (closes #54)

### LINKED ISSUE #54: cv with drupal7 multisite
I've been playing around with cv in a drupal7 multisite setup, and I've run into two related problems:

1) The only bootstrap levels that work are classloader/settings; booting drupal will fail (at the http redirect in includes/install.inc:install_goto()). It works if I set $_SERVER[HTTP_HOST] = 'mysite.org' in Bootstrap.php+CmsBootstrap.php, before they simulate the web environment. (Is there a way to detect this kind of redirect failure? Right now it fails silently which is not ideal.)

2) cv's search for civicrm.settings.php does not find anything. I'd expect cv to find the settings file if I first cd into [cmsroot]/sites/mysite (or one of its children), similarly to how it works with drush. But findDrupalDirs() tries to guess folder names under sites/ based on $options['httpHost'] (pulled from $_SERVER[HTTP_HOST]) which is empty. So it ends up checking just [cmsroot]/sites/ and [cmsroot]/sites/default.

So it seems like cv needs a "--hostname" option, that would be used to set $_SERVER[HTTP_HOST]. This would be analogous to drush's --uri option, and would address both problems.

I also wonder if the settings file search procedure should be changed to make findCivicrmSettingsPhp() also check all ancestor directories (starting with cwd) for the settings file. This will help with drupal's sites.php aliasing, where the site directory name and host name might be different.

Does something along these lines seem reasonable? You can see a preliminary version of it here: https://github.com/pcurrier/cv/commits/master

pcurrier:
To close the loop on issue #54 (problems with drupal7 multisite):

* add --hostname parameter for setting $_SERVER[HTTP_HOST] (needed to boot drupal; otherwise won't find the drupal settings file)
* make use of any directory mappings in drupal's sites.php (helps cv find civicrm.settings.php)

With these two changes, cv works well on drupal7 multisite.

### DIFF
--- a/src/Bootstrap.php
+++ b/src/Bootstrap.php
@@ -382,10 +382,18 @@ class Bootstrap {
    * @return array
    */
   protected function findDrupalDirs($cmsRoot) {
+    $sites = array();
+    if (file_exists("$cmsRoot/sites/sites.php")) {
+      include("$cmsRoot/sites/sites.php");
+    }
     $dirs = array();
     $server = explode('.', implode('.', array_reverse(explode(':', rtrim($this->options['httpHost'], '.')))));
     for ($j = count($server); $j > 0; $j--) {
-      $dirs[] = "$cmsRoot/sites/" . implode('.', array_slice($server, -$j));
+      $s = implode('.', array_slice($server, -$j));
+      if (isset($sites[$s]) && file_exists("$cmsRoot/sites/" . $sites[$s])) {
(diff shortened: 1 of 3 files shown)


totten:
Thanks @pcurrier for taking the plunge and pushing this forward!

I've expanded on this a bit with #58 and will comment more on #54.

PR #58, opened 2020-01-09
Multiple updates for drupal7 multisite (closes #54)

### LINKED ISSUE #54: cv with drupal7 multisite

totten:
This is a continuation of #57. A few additions/changes:

* Both `Bootstrap` and `CmsBootstrap` protocols should accept the `httpHost` option. (Previously, only `CmsBootstrap` accepted it.) The `httpHost` should be passed in as a documented option.
* Add some options for multisite usage in which you don't have to explicitly pass `--hostname`:
    * For `Bootstrap`, check if the PWD indicates the preferred site.
    * For `CmsBootstrap`, when using the environment `CIVICRM_BOOT`, look for a parameter `?host=<HOSTNAME>`.
        * (It would also be nice to fallback to checking PWD or somesuch. Intuitively you might look for `/var/www/.../sites/FOO.BAR` and then say `httpHost=FOO.BAR`. However, the `conf_path()` algorithm is a bit rich, and I'm not sure that heuristic is particularly accurate. Better to let that idea marinade a bit more...)
* Make it more realistic to use CMS-first protocol on a day-to-day basis. If you specify boot options in `CIVICRM_BOOT`, then most commands will now respect that. But if you don't, then it continues using the traditional protocol.

For testing, I setup a local multisite with two folders (`sites/default` and `sites/foo.dmaster.bknix`) and ran several permutations of `cv url civicrm/admin`:

```
for BASE in "$HOME/bknix/build/dmaster/web/"{,sites/default,sites/foo.dmaster.bknix} ; do cd "$BASE"; for LVL in full cms-full ; do for OPT in '' '--hostname=foo.dmaster.bknix' '--hostname=dmaster.bknix' ; do echo -n "PWD=$PWD; LEVEL=$LVL OPT=$OPT ... Output: " ; ~/src/cv/bin/cv url civicrm/admin --level=$LVL $OPT ; done ; done ; done
```

Which outputs (*very wide text*):

```
PWD=~/bknix/build/dmaster/web; LEVEL=full OPT=                                                          Output: "http://dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web; LEVEL=full OPT=--hostname=foo.dmaster.bknix                              Output: "http://foo.dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web; LEVEL=full OPT=--hostname=dmaster.bknix                                  Output: "http://dmaster.bknix:8001/civicrm/admin"

PWD=~/bknix/build/dmaster/web; LEVEL=cms-full OPT=                                                      Output: "http://dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web; LEVEL=cms-full OPT=--hostname=foo.dmaster.bknix                          Output: "http://foo.dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web; LEVEL=cms-full OPT=--hostname=dmaster.bknix                              Output: "http://dmaster.bknix:8001/civicrm/admin"

PWD=~/bknix/build/dmaster/web/sites/default; LEVEL=full OPT=                                            Output: "http://dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web/sites/default; LEVEL=full OPT=--hostname=foo.dmaster.bknix                Output: "http://foo.dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web/sites/default; LEVEL=full OPT=--hostname=dmaster.bknix                    Output: "http://dmaster.bknix:8001/civicrm/admin"

PWD=~/bknix/build/dmaster/web/sites/default; LEVEL=cms-full OPT=                                        Output: "http://dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web/sites/default; LEVEL=cms-full OPT=--hostname=foo.dmaster.bknix            Output: "http://foo.dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web/sites/default; LEVEL=cms-full OPT=--hostname=dmaster.bknix                Output: "http://dmaster.bknix:8001/civicrm/admin"

PWD=~/bknix/build/dmaster/web/sites/foo.dmaster.bknix; LEVEL=full OPT=                                  Output: "http://foo.dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web/sites/foo.dmaster.bknix; LEVEL=full OPT=--hostname=foo.dmaster.bknix      Output: "http://foo.dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web/sites/foo.dmaster.bknix; LEVEL=full OPT=--hostname=dmaster.bknix          Output: "http://dmaster.bknix:8001/civicrm/admin"

PWD=~/bknix/build/dmaster/web/sites/foo.dmaster.bknix; LEVEL=cms-full OPT=                              Output: "http://dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web/sites/foo.dmaster.bknix; LEVEL=cms-full OPT=--hostname=foo.dmaster.bknix  Output: "http://foo.dmaster.bknix:8001/civicrm/admin"
PWD=~/bknix/build/dmaster/web/sites/foo.dmaster.bknix; LEVEL=cms-full OPT=--hostname=dmaster.bknix      Output: "http://dmaster.bknix:8001/civicrm/admin"

```

All of these look correct except for:

```
PWD=~/bknix/build/dmaster/web/sites/foo.dmaster.bknix; LEVEL=cms-full OPT=                              Output: "http://dmaster.bknix:8001/civicrm/admin"
```

which is because `CmsBootstrap` doesn't support multisite detection from the PWD.

### DIFF
--- a/src/Bootstrap.php
+++ b/src/Bootstrap.php
@@ -382,10 +382,18 @@ class Bootstrap {
    * @return array
    */
   protected function findDrupalDirs($cmsRoot) {
+    $sites = array();
+    if (file_exists("$cmsRoot/sites/sites.php")) {
+      include("$cmsRoot/sites/sites.php");
+    }
     $dirs = array();
     $server = explode('.', implode('.', array_reverse(explode(':', rtrim($this->options['httpHost'], '.')))));
     for ($j = count($server); $j > 0; $j--) {
-      $dirs[] = "$cmsRoot/sites/" . implode('.', array_slice($server, -$j));
+      $s = implode('.', array_slice($server, -$j));
+      if (isset($sites[$s]) && file_exists("$cmsRoot/sites/" . $sites[$s])) {
+        $s = $sites[$s];
+      }
+      $dirs[] = "$cmsRoot/sites/$s";
     }
     $dirs[] = "$cmsRoot/sites/default";
     return $dirs;
--- a/src/Bootstrap.php
+++ b/src/Bootstrap.php
@@ -63,6 +63,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  *   - env: string|NULL. The environment variable which may contain the path to
  *     civicrm.settings.php. Set NULL to disable.
  *     (Default: CIVICRM_SETTINGS)
+ *   - httpHost: string|NULL. For multisite, the HTTP hostname.
  *   - prefetch: bool. Whether to load various caches.
  *     (Default: TRUE)
  *   - settingsFile: string|NULL. The full path to the civicrm.settings.php
--- a/src/Bootstrap.php
+++ b/src/Bootstrap.php
@@ -340,13 +340,13 @@ class Bootstrap {
     switch ($cmsType) {
       case 'backdrop':
         $settings = $this->findFirstFile(
-           array_merge($this->findDrupalDirs($cmsRoot), array($cmsRoot)),
+           array_merge($this->findDrupalDirs($cmsRoot, $searchDir), array($cmsRoot)),
           'civicrm.settings.php'
         );
         break;
 
       case 'drupal':
-        $settings = $this->findFirstFile($this->findDrupalDirs($cmsRoot), 'civicrm.settings.php');
+        $settings = $this->findFirstFile($this->findDrupalDirs($cmsRoot, $searchDir), 'civicrm.settings.php');
         break;
 
       case 'joomla':
@@ -379,9 +379,20 @@ class Bootstrap {
    *
    * @param string $cmsRoot
    *   The root of the Drupal installation.
+   * @param string $searchDir
+   *   The location from which we're searching. Usually PWD.
    * @return array
    */
-  protected function findDrupalDirs($cmsRoot) {
+  protected function findDrupalDirs($cmsRoot, $searchDir) {
+    // If there's no explicit host and we start the search from "web/sites/FOO/...", then infer subsite path.
+    $sitesRoot = "$cmsRoot/sites";
+    $sitesRootQt = preg_quote($sitesRoot, ';');
+    if (empty($this->options['httpHost']) && preg_match(";^($sitesRootQt/[^/]+);", $searchDir, $m)) {
+      if (basename($m[1]) !== 'all') {
+        return [$m[1]];
+      }
+    }
+
     $sites = array();
     if (file_exists("$cmsRoot/sites/sites.php")) {
       include("$cmsRoot/sites/sites.php");
--- a/src/Bootstrap.php
+++ b/src/Bootstrap.php
@@ -208,6 +208,10 @@ class Bootstrap {
         $_SERVER['REMOTE_ADDR'] = "127.0.0.1";
         $_SERVER['SERVER_SOFTWARE'] = NULL;
         $_SERVER['REQUEST_METHOD'] = 'GET';
+        if (!empty($options['httpHost'])) {
+          // Hint for D7 multisite
+          $_SERVER['HTTP_HOST'] = $options['httpHost'];
+        }
         if (ord($_SERVER['SCRIPT_NAME']) != 47) {
           $_SERVER['SCRIPT_NAME'] = '/' . $_SERVER['SCRIPT_NAME'];
         }
--- a/src/CmsBootstrap.php
+++ b/src/CmsBootstrap.php
@@ -46,8 +46,9 @@ class CmsBootstrap {
   protected $options = array();
 
   /**
-   * Symphony OutputInterface provide during booting to enable the command-line
-   * 'v', 'vv' and 'vvv' options
+   * @var \Symfony\Component\Console\Output\OutputInterface
+   *   Provide during booting to enable the command-line verbosity
+   *   options ('v', 'vv' and 'vvv').
    * @see http://symfony.com/doc/current/console/verbosity.html
    */
   protected $output = FALSE;
@@ -112,6 +113,12 @@ class CmsBootstrap {
       if (!isset($this->options['user']) && parse_url($cmsExpr, PHP_URL_USER)) {
         $this->options['user'] = parse_url($cmsExpr, PHP_URL_USER);
       }
+      if (parse_url($cmsExpr, PHP_URL_QUERY)) {
+        parse_str(parse_url($cmsExpr, PHP_URL_QUERY), $query);
+        if (!empty($query['host'])) {
+          $this->options['httpHost'] = $query['host'];
+        }
+      }
     }
     else {
       $this->writeln("Find CMS...", OutputInterface::VERBOSITY_DEBUG);
@@ -120,7 +127,8 @@ class CmsBootstrap {
 
     $this->writeln("CMS: " . Encoder::encode($cms, 'json-pretty'), OutputInterface::VERBOSITY_DEBUG);
     if (empty($cms['path']) || empty($cms['type']) || !file_exists($cms['path'])) {
-      throw new \Exception("Failed to parse or find a CMS");
+      $cmsJson = json_encode($cms, JSON_UNESCAPED_SLASHES);
+      throw new \Exception("Failed to parse or find a CMS $cmsJson");
     }
 
     if (PHP_SAPI === "cli") {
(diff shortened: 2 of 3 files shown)


pcurrier:
One question I have now that I understand CIVICRM_BOOT a little better: would it make sense to specify the variable as "Drupal://foo.example.org/path" rather than "Drupal://path?host=foo.example.org"? Then you'd use parse_url() with PHP_URL_HOST/PHP_URL_PATH to extract the host and path (rather than concatenating them together to form the path, as it does now). Might be less surprising to users if the hostname appears in the same place as it appears in other analogous URLs.

totten:
Yeah, after a day or so, agree the `?host` bit is feeling slightly weird. As I was reading about the other ways multisite can work (ports and subdirs), and thinking abut the goal of generating valid URLs (which would also involve http-vs-https), I started to think that maybe the spec should have a base URL. Which could be:

```bash
CIVICRM_BOOT="Drupal://var/www?baseUrl=https%3A%2F%2Ffoo.example.org%3A8001"
```

But that's a lot of symbols. I wonder if maybe it'd make more sense as two parts -- the boot/loader spec and the web spec., e.g.

```bash
CIVICRM_BOOT="Drupal://var/www https://foo.example.org:8001"
## Note: Spaces are not valid in URLs, so they can be a delimiter.
```

The first term (`Drupal://var/www`) gives some CMS-specific bits for how to boot. The second term (`https://foo.example.org:8001`) tells it how to simulate the web environment.

nganivet:
@totten Seems strange to try to pack 3 different concepts in the same variable, why simply not use multiple variables? How about for example CIVICRM_UF=Drupal; CIVICRM_UF_DIR=/var/www; CIVICRM_UF_URL=https://foo.example.org:8001? Must more legible, and much closer to what we already have in civicrm.settings.php.

totten:
Yeah... there are merits in both packing those bits together and splitting them apart.

* __Splitting__:
    * If you're good with config-mgmt but don't often use this specific tooling/subsystem, then splitting the bits apart makes it easier to read.
    * If you need to write some script which loops and alters one thing (e.g. the base URL in a multisite) while holding the other parts constant (e.g. the CMS type), then splitting will look cleaner.
    * There are cases where env-var `CIVICRM_UF` maps into PHP constant `CIVICRM_UF`. I like that symmetry. (Tho it's a mixed bag - I don't think the symmetry would be quite so elegant with the other bits.)
* __Packing__:
    * If you're reading docs or making a support request, it's easier to communicate about one variable than to make sure you communicate the right subset of 3-5 variables.
    * If you're not comfortable with the configuration-management (e.g. combined env-vars of `~/.bashrc` vs `/etc/profile` vs `Jenkinsfile` vs `.gitlab.yml`), then packing the data into one expression makes it less error-prone.
   * I'm somewhat in awe of how effective URLs have been at packing different fields -- scheme, hostname, port, path, query-string, etc -- into the dataflows/experiences of web-browsers, HTML, etc.

If the main use-case here was (e.g.) one person managing many different permutations with a specific configuration-management system, then I'd probably do a hard-lean toward splitting env-vars (b/c the different vars could look nicer in the config). But I think cv's situation is more like parachuting into an "organic" stew of environments/people/toolchains/skillsets/communication-processes, where I'm more worried that you'd get misunderstandings because folks subjectively elide/omit variables from discussion.

Am I speaking sensibly, or is that crazy talk?

nganivet:
@totten no offense but it sounds a bit like crazy talk to me: the KISS principle promotes using one variable per information, and I have never seen the Drupal:// scheme before, or anyone packing multiple concepts with spaces inside the same environment variable. But I have no skin in this game, so do as you see best.

MegaphoneJon:
I agree with Nicolas, but it's more important that this function exist than the implementation particulars.

What can be done to advance this?  Does it need to be tested?

nganivet:
Reading my comments again I persist in saying that packing is a bad idea that will bite us afterwards, but again I have no skin in the game.

@totten that said, I'd be curious to see where this idea came from: do you have other examples of 'packing' as such?
Also, why go through the trouble of 'packing' this variable if the first thing you are going to do in your code is 'unpack' it to consume the information it encodes.

MegaphoneJon:
I attempted to use this patch on D8 today, but the CMS-finding function is looking for a file that doesn't exist on D8.

This is more important for D8 because there's no `drush cvapi` command and `cron.php` isn't multisite-aware.  I'm going to try tomorrow to get AuthX working with a REST endpoint in hopes that this will trigger a secondary site's cron, but if it doesn't there may be no cron option for D8 w/ Domain Access.

mlutfy:
hmm maybe I missed a bit, but I tried:

```
cd /tmp/
git clone https://github.com/MegaphoneJon/cv/
git checkout d8-multisite
composer install
```

then:

```
cd /path/to/my/platform/sites/foo.example.org
/tmp/cv/bin/cv --hostname="foo.example.org" url
# -> "The specified database connection is not defined: default"

/tmp/cv/bin/cv --httpHost="foo.example.org" url
# -> The "--httpHost" option does not exist
```

Luciano mentioned they had a fork for Aegir, so I tested that:
https://github.com/ixiam/cv/commits/aegir

and that weirdly works without any options:

```
$ cd /path/to/my/platform/sites/foo.example.org
$ /tmp/cv/bin/cv url
"https://foo.example.org/"
```

(I tested with another site in the same platform, and it also gave the right output)

however, their fork is a few months behind master.

MegaphoneJon:
@mlutfy Here's what I'm using in production:
```
/home/jon/cv/bin/cv --hostname="california.crm.mysite.org" --cwd=/var/www/crm.mysite.org/web --user=cron --quiet api job.execute
```

mlutfy:
Yikes time flies. I did some more testing and figured why I was having issues, and was Aegir-related (db credentials must not be "cloaked"). This PR worked for me.

Tested with:

```
cd /var/aegir/platforms/civicrm-d8/sites/crm.example.org
/home/foo/tmp/cv --hostname="crm.example.org" api3 contact.get
```

There is a conflict with code comments though.

@totten Still interested in merging this? It would really help in my hosting environments.

sluc23:
I didn't know why i didn't see this post until today :S !
as @mlutfy said, we had our own fork of `cv` working in D7/D9 multisite, running under an *AEgir* environment and based on `drush` commands, without any extra variable.. it auto-discovers the site according to the path where it's called (with some tweeking).. maybe too custom for this PR, but just in case I leave here the latest version, updated until current `v0.3.16` tag

https://github.com/civicrm/cv/compare/v0.3.16...ixiam:v0.3.16-AEgir?expand=1

totten:
The replacement PR #124 was previously merged.

PR #59, opened 2020-02-03
Address php 7.4 deprecation

eileenmcnaughton:
Deprecated: Array and string offset access syntax with curly braces is deprecated 

https://stackoverflow.com/questions/59158548/array-and-string-offset-access-syntax-with-curly-braces-is-deprecated


eileenmcnaughton:
@totten  I think these are no-brainers - https://github.com/civicrm/cv/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aopen+7.4

totten:
Close in favor of #64 (which covers all 4 of the `{0}` cases).

PR #60, opened 2020-02-03
Address php 7.4 deprecation notice

eileenmcnaughton:
https://stackoverflow.com/questions/59158548/array-and-string-offset-access-syntax-with-curly-braces-is-deprecated


totten:
Close in favor of #64 (which covers all 4 of the `{0}` cases).

PR #62, opened 2020-02-10
quote MySQL password

MegaphoneJon:
If you have globbing characters (e.g. `*`) in your MySQL password, `cv` will fail to connect.  Quoting the password in the generated `defaults-file` fixes the problem and seems to have no ill effect.

### DIFF
--- a/src/Util/Datasource.php
+++ b/src/Util/Datasource.php
@@ -137,7 +137,7 @@ class Datasource {
     $data = "[client]\n";
     $data .= "host={$this->host}\n";
     $data .= "user={$this->username}\n";
-    $data .= "password={$this->password}\n";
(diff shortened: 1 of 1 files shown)


src/Util/Datasource.php
```diff
@@ -137,7 +137,7 @@ public function toMySQLArguments($tmpDir) {
     $data = "[client]\n";
     $data .= "host={$this->host}\n";
     $data .= "user={$this->username}\n";
-    $data .= "password={$this->password}\n";
+    $data .= "password='{$this->password}'\n";
```
totten: Good idea.

I suppose one could have a password with a quote mark?  Do you happen to know what escaping rules they use MySQL's file-format?  Maybe `addslashes()` or `addcslashes()` would come closer?

PR #64, opened 2020-03-06
Change deprecated syntax for string offsets

MegaphoneJon:
See here: https://wiki.php.net/rfc/deprecate_curly_braces_array_access

### DIFF
--- a/src/BuildkitReader.php
+++ b/src/BuildkitReader.php
@@ -35,7 +35,7 @@ class BuildkitReader {
     $lines = explode("\n", file_get_contents($shFile));
     $result = array();
     foreach ($lines as $line) {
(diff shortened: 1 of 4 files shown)


totten:
Ran the test suite on my local with PHP 7.4 - confirmed that this fixes some tests that were failing due to the deprecation notice. The `cv` test suite now passes on 7.4.

Did a grep for other expressions that might be string-offsets (e.g. grep for `'\{[0-9]\}'` and `'\{\$'`) and couldn't find anything that would affect the operation of `cv`. The matches all seemed to be dynamic properties (`$obj->{$prop}`) and string interpolation (`"string with {$var}"`), but those aren't string offsets.

totten:
Published in v0.3.2.

PR #71, opened 2020-03-27
(POC) Update symfony/console 2.x => 3.x

totten:
This patch updates `symfony/console`.  A straight update, however, causes a regression: if you run `php:eval` or `php:script` with an invalid script, it will no longer have a semantically appropriate error code. Also, the formatting of the error seems less appropriate.

This patch addresses the exit code (which has more functional impact on scripting and unit-tests), but I'm kind of surprised that `symfony/console` went from (v2) reporting the error well to (v3) not reporting the error as well.  It makes me think that maybe they changed the design and have an alternative in mind.

There's not much urgency to updating this AFAIK, so I think it's ok to marinate a bit in case one finds time to investigate why `symfony/console` changed.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -231,16 +231,16 @@
         },
         {
             "name": "psr/log",
-            "version": "1.0.2",
+            "version": "1.1.3",
             "source": {
                 "type": "git",
                 "url": "https://github.com/php-fig/log.git",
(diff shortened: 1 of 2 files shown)


totten:
Replaced by #106

PR #78, opened 2020-08-04
cv upgrade:db - Fix multilingual support by flagging upgrade-mode

seamuslee001:
…rade to ensure that the correct schema is retrieved

ping @totten 

You can test this by loading up CiviCRM 5.13.x making it a multilingual instance and then using cv to upgrade the database without this it borks

### DIFF
--- a/src/Command/UpgradeDbCommand.php
+++ b/src/Command/UpgradeDbCommand.php
@@ -143,6 +143,13 @@ Examples:
 
     $output->writeln("<info>Upgrade to <comment>$codeVer</comment> completed.</info>", $niceMsgVerbosity);
 
+    if (version_compare($codeVer, '5.26.alpha', '<')) {
+      // Work-around for bugs like dev/core#1713.
+      // Note that #1713 didn't affect earlier versions of `cv` because they mistakenly omitted CIVICRM_UPGRADE_ACTIVE.
+      $output->writeln('<info>Detected CiviCRM 5.25 or earlier. Force flush.</info>');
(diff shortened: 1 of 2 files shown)


totten:
Aaah, nice catch. Agree this is more compliant with the protocol for an upgrader, and it'll make `cv` behave more like all the other upgraders.

Interestingly, I noticed the same discrepancy while testing the upgrader/menu-rebuild bug (https://lab.civicrm.org/dev/core/-/issues/1713 - fixed in 5.26). Basically, the `cv` upgrader was getting a good outcome, but the normal upgrader got a bad outcome, and they differed because... `cv` forgot to set `CIVICRM_UPGRADE_ACTIVE`. 

However, as you've found here, the non-compliance means that there are *other* bugs - ie a multilingual bug.

What this means... for Civi 5.25 or earlier, toggling `CIVICRM_UPGRADE_ACTIVE` in `cv` would simply trade bugs. (One bug affects all sites, but it has a softer/recoverable problem. The other bug affects only multilingual sites, but it causes a harder failure.)

Of course, on Civi 5.26+, setting `CIVICRM_UPGRADE_ACTIVE` is straight-up improvement.

I'm a little torn about how much to care about <=5.25. On one hand, 5.27 is going into ESR now, so 5.25 feels kinda dead-end. On the other hand, 5.25 is only 2 months old, and we may still need to run intermediate upgrades for debugging/triage purposes.

So I've added a commit to address those versions. If you're fine with the new commit @seamuslee001 , then let's merge.

seamuslee001:
Looks fine to me @totten

PR #85, opened 2020-11-21
Allow civicrm/composer-downloads-plugin v3 components.

kewljuice:


totten:
Looks good to me. thanks @kewljuice

PR #89, opened 2021-02-15
Update example in README for getPath

demeritcowboy:
Current example of get() gives an error.


MikeyMJCO:
If you add an example of `getUrl()` as well I can close #90

Issue #95, opened 2021-03-25
Deprecation warnings breaking php:boot PHP code


artfulrobot:
Locally I'm running cv 0.3.6 on a docker container with php 7.4

PHP 7.4 causes every extension that has not recently regenerated its civix files to issue deprecation errors like:

```
Deprecated: Array and string offset access syntax with curly braces is deprecated in /var/www/.../sites/all/civicrm_extensions/uk.co.mjwconsult.checksum/checksum.civix.php on line 169

Deprecated: Array and string offset access syntax with curly braces is deprecated in /var/www/../sites/all/civicrm_extensions/uk.co.vedaconsulting.mosaico/mosaico.civix.php on line 246

Deprecated: Array and string offset access syntax with curly braces is deprecated in /var/www/.../sites/all/civicrm_extensions/org.civicrm.recentmenu/recentmenu.civix.php on line 246

... and the rest ...
```

While those are useful to know about, the problem is that they are output on STDOUT not STDERR, and as such they are included in `$result` and as such code like the [example code in the README](https://github.com/civicrm/cv/blame/master/README.md#L120) which looks for `/**BEGINPHP*/` won't find it, because it's looking at `Deprecated:...`.

artfulrobot:
I wonder whether, since this is a command line utility, we should include

```php
<?php
ini_set ( 'display_errors' , 'stderr' );
```

which would ([docs](https://www.php.net/errorfunc.configuration#ini.display-errors)) send the errors to stderr?

PR #96, opened 2021-03-25
Fix for issue 95: use STDERR for display_errors

artfulrobot:
This swaps out setting `display_errors = 1` for `display_errors = 'stderr'` as documented to mean, unsurprisingly, that errors get pumped out on STDERR.

This means command line users can get rid of errors easily, e.g. for the sake of being able to  boot civi - see #95 

I've tried to run the unit tests, but they're hanging when they get to a certain place, bu thtat happens before this change too?

### DIFF
--- a/bin/cv
+++ b/bin/cv
@@ -1,6 +1,6 @@
 #!/usr/bin/env php
 <?php
-ini_set('display_errors', 1);
+ini_set('display_errors', 'stderr');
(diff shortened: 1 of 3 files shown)


totten:
Yeah, this change looks pretty sensible. (I didn't know that `display_errors=stderr` was an option!)

I was able to run the test suite on top of it -- I suspect the hang might've been something else.

Merging.

PR #108, opened 2021-12-23
Fix wrong variable in cmsbootstrap

demeritcowboy:
Should be self-explanatory.


totten:
Yup, agree with this. Also gave it light `r-run`, and it seemed good.

demeritcowboy:
Thanks!

Issue #117, opened 2022-04-06
Symfony cmf removed from drupal


demeritcowboy:
See https://www.drupal.org/node/3151009.  In bootDrupal8() in src/cmsBootstrap.php, this sort of works:

```diff
-   $request->attributes->set(RouteObjectInterface::ROUTE_OBJECT, new Route('<none>'));
-   $request->attributes->set(RouteObjectInterface::ROUTE_NAME, '<none>');
+   $request->attributes->set(\Drupal\Core\Routing\RouteObjectInterface::ROUTE_OBJECT, new Route('<none>'));
+   $request->attributes->set(\Drupal\Core\Routing\RouteObjectInterface::ROUTE_NAME, '<none>');
```

I say "sort of" because it seems weird that it compiles still with the `use` statement still at the top for Symfony\Cmf\Component\Routing\RouteObjectInterface. But if php doesn't care then :shrug:.

Without the patch trying to use cv with drupal 10 you get `Error: Class "Symfony\Cmf\Component\Routing\RouteObjectInterface" not found in ...\src\CmsBootstrap.php on line 259 #0 [internal function]: Civi\Cv\CmsBootstrap->bootDrupal8()`

I haven't tested the patch with drupal 8 but I don't know how much support there should be for drupal 8 anymore.

PR #120, opened 2022-04-18
#117 - Symfony cmf removed in drupal 10

demeritcowboy:
Addresses https://github.com/civicrm/cv/issues/117

This patch works in drupal 9 and 10 but doesn't work in drupal 8, but drupal 8 been eol for 6 months and doesn't get security fixes and civi can be run equally as well on drupal 9.

Before
-------
Simple example from within a drupal 10 install (don't try to install civi):
`cv ev --level=cms-only "echo Drupal::VERSION;"`

`Error: Class "Symfony\Cmf\Component\Routing\RouteObjectInterface" not found in ...\src\CmsBootstrap.php on line 259 #0 [internal function]: Civi\Cv\CmsBootstrap->bootDrupal8()`

After
-----
9.3.9 (or whatever)

### DIFF
--- a/src/CmsBootstrap.php
+++ b/src/CmsBootstrap.php
@@ -2,7 +2,6 @@
 namespace Civi\Cv;
 
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Cmf\Component\Routing\RouteObjectInterface;
 use Symfony\Component\Routing\Route;
 
 /**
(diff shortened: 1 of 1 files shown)


totten:
Ported the revision from `civicrm-core` PR.

Tested on D8 and confirmed that it still works with the adjusted conditional. I don't have D10 to `r-run`, but the change sounds plausible enough.

demeritcowboy:
Hmm porting the part from core that keeps drupal 8 support doesn't actually work here in bootstrap because it looks like class_exists is always false since it hasn't got that far yet. I'll put up a PR to remove that.

PR #124, opened 2022-05-16
Drupal multisite support

MegaphoneJon:
This supersedes #58 - it's largely the same but incorporates the changes I describe in https://github.com/civicrm/cv/pull/58#issuecomment-905822794.  @mlutfy confirms this works for him, I realized I never submitted a PR with my changes.

@pcurrier @sluc23 you both expressed interest in this.

### DIFF
--- a/src/Bootstrap.php
+++ b/src/Bootstrap.php
@@ -63,6 +63,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  *   - env: string|NULL. The environment variable which may contain the path to
  *     civicrm.settings.php. Set NULL to disable.
  *     (Default: CIVICRM_SETTINGS)
+ *   - httpHost: string|NULL. For multisite, the HTTP hostname.
  *   - prefetch: bool. Whether to load various caches.
  *     (Default: TRUE)
  *   - settingsFile: string|NULL. The full path to the civicrm.settings.php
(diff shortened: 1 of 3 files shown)


sluc23:
looks good @MegaphoneJon .. but i won't be a regular user of this one.. we are using iXiam's fork that relies on `drush8` without having to add an extra parameter. It's a very tweaked version, but it fits for our needs.
Anyway i'll follow up closely this issue to know how it moves forwards!

totten:
Since you've resolved the conflicts, and it seems to have worked for a few people in different times/ways, I think we should go ahead and merge. The revisions on top of #58 also look good (eg more logging).

There is a wrinkle wrt consistency between commands:

* There are a couple commands (`core:install`, `core:check-req`, and `core:uninstall`) which have the option `--cms-base-url`. That wasn't really about multisite (*it was about `CIVICRM_UF_BASEURL`*), but it is roughly the same content as `--hostname`.
* It's nice that `--hostname` doesn't require the full `http://` stuff.
* It's nice that `--cms-base-url` is more expressive (ie you can specify protocol+port).
* It would be nice if `--hostname` and `--cms-base-url` were shorter. (`-H`, `-l`, or `--url` seem more usable...)
*  IIRC, `drush`'s `-l` accepts mixed notations (eg `-l example.com`, `-l https://example.com`, `-l example.com:8000`, `-l https://example.com:8443`). I'm not sure about `wp-cli'`s `--url`. ([This page](https://danielbachhuber.com/tip/run-wp-cli-command-wordpress-multisite/) suggests it is a hostname, and I vaguely recall that `--url` was a misnomer.)

I think a decent play here would be:

* Go ahead with adding this option
* In another PR, work on consolidating `--hostname` and `--cms-base-url` to `-l`/`--url`. Do a little extra parsing to accept mixed notations. Maybe bump the version to `0.4` and give a deprecation warning for one of the options.

Flagging "merge ready", since I think it's "better in than out", but if anyone says we should wait to reconcile these things first, we can hold it a bit longer.

PR #126, opened 2022-05-19
Remove drupal 8 support since using class_exists doesn't work properly in cmsbootstrap

demeritcowboy:
It looks like class_exists is always false so while it seems like it works in drupal 8 and 9 because the old class exists anyway, in drupal 10 it fails.

So just remove drupal 8 support.

### DIFF
--- a/src/CmsBootstrap.php
+++ b/src/CmsBootstrap.php
@@ -254,11 +254,8 @@ class CmsBootstrap {
     $kernel->preHandle($request);
     $container = $kernel->rebuildContainer();
     // Add our request to the stack and route context.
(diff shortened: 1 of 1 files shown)


totten:
Thanks @demeritcowboy.

Yeah, I'm fine with that.

If anyone really needs `cv` on D8 for the next couple months, they can download an older version. No need to keep carrying it here.

demeritcowboy:
Thanks!

In hindsight my PR description was a bit light. For anyone needing more context see

* The previous PR: https://github.com/civicrm/cv/pull/120
* https://lab.civicrm.org/dev/drupal/-/issues/178
* https://lab.civicrm.org/dev/drupal/-/issues/176

PR #132, opened 2022-09-14
CmsBootstrap::bootDrupal8(): bootstrap Drupal 9 with --user config

### LINKED ISSUE #113: Drupal 9 compatibility
I tried to install a Drupal 9-based CiviCRM, but it failed:

```
cv core:install --no-interaction --cwd=/civi/dir --lang=en_GB --cms-base-url=https://crm.dev --user=CRM_Admin

Error: Call to undefined method Drupal::entityManager() in phar:///usr/local/bin/cv/src/CmsBootstrap.php on line 269 #0 [internal function]: Civi\Cv\CmsBootstrap->bootDrupal8()
```

`Drupal::entityManager()` was removed in Drupal 9. https://www.drupal.org/project/config_entity_revisions/issues/3196948

So I'm wondering, do `cv` support Drupal 9?

I've a quick-fix which seems to work, and ready to submit a PR, I'm just not sure that it's the best way from an architectural point of view.

semseysandor:
fix #113 

Adaptive logic added to check if `\Drupal::entityManager()` can be called (D8), if not it calls `\Drupal::entityTypeManager()` and gets users the D9 way.

Note: for some reason (I can't see why) `is_callable` returned `true` for the non-existent method and resulted in an "Call to undefined method" whereas `method_exists` returned `false` correctly.

### DIFF
--- a/src/CmsBootstrap.php
+++ b/src/CmsBootstrap.php
@@ -272,11 +272,21 @@ class CmsBootstrap {
     }
 
     if ($cmsUser) {
-      $entity_manager = \Drupal::entityManager();
-      $users = $entity_manager->getStorage($entity_manager->getEntityTypeFromClass('Drupal\user\Entity\User'))
-        ->loadByProperties(array(
-          'name' => $cmsUser,
-        ));
+      // Drupal 8
+      if (method_exists('Drupal', 'entityManager')) {
+        $entity_manager = \Drupal::entityManager();
+        $users = $entity_manager->getStorage($entity_manager->getEntityTypeFromClass('Drupal\user\Entity\User'))
+          ->loadByProperties(array(
+            'name' => $cmsUser,
+          ));
+      } else {
+        // Drupal 9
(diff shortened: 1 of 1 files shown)


totten:
Thanks for posting this @semseysandor. I agree that this looks consistent with how the contract evolved from D8/D9.

For testing, I didn't do a full `core:install`, but I was able to demonstrate the improvement with:

Before:
```
[bknix-min:~/bknix/build/drupal9-clean] ~/src/cv/bin/cv ev --level='cms-full' 'echo 123;' --user=admin
Error: Call to undefined method Drupal::entityManager() in /Users/totten/src/cv/src/CmsBootstrap.php on line 275 #0 [internal function]: Civi\Cv\CmsBootstrap->bootDrupal8('/Users/totten/b...', 'admin', 'localhost')
```

After:

```
[bknix-min:~/bknix/build/drupal9-clean] ~/src/cv/bin/cv ev --level='cms-full' 'echo 123;' --user=admin
123
```

PR #140, opened 2022-12-27
PHP8 breaks passing blank values in api4

MegaphoneJon:
Previously, I used this to fix AuthX for sites behind basic auth:
```shell
cv api4 Setting.set +v authx_header_cred=
```

However, in PHP8, this throws an error:
```
In AbstractPlusParser.php line 68:
                         
  Failed to parse JSON:  
```

That is because in PHP8, [every string is considered to have an empty string at every position in a string](https://php.watch/versions/8.0/string-function-empty-needles).

This checks explicitly for an empty string.

That said - aren't `needle` and `haystack` reversed here?

### DIFF
--- a/src/Util/AbstractPlusParser.php
+++ b/src/Util/AbstractPlusParser.php
@@ -75,7 +75,7 @@ abstract class AbstractPlusParser {
    * @return mixed
    */
   protected function parseValueExpr($expr) {
-    if (strpos('{["\'', $expr[0]) !== FALSE) {
+    if ($expr !== '' && strpos('{["\'', $expr[0]) !== FALSE) {
       return $this->parseJsonNoisily($expr);
     }
     else {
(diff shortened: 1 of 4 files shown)


totten:
Confirmed -- this makes PHP8 behavior match PHP7. 👍 Added another related fix+test. Merging.

> ```php
> strpos('{["\'', $expr[0]) !== FALSE
> ```
>
> That said - aren't needle and haystack reversed here?

No, but the conditional here may be unusual. It determines whether `$expr[0]` (*ie the first character*) appears in a list of JS/JSON meta-characters (`{`, `[`, `"`, `'`). So it matches a JSON object, JSON array, or JSON string.

> ```
> cv api4 Setting.set +v authx_header_cred=
> ```

Aside/fun fact/bike-shed -- I'd probably do:

```
cv vset authx_header_cred=[]
```

`vset` and `vget` are just a little richer. (Ex: try `cv vget /authx/` or `cv vget -v authx_header_cred`)

PR #142, opened 2023-02-03
Initial support for "Standalone" deployments

artfulrobot:
This PR includes additions to cv to support Civi Standalone

### DIFF
--- a/src/CmsBootstrap.php
+++ b/src/CmsBootstrap.php
@@ -410,7 +410,9 @@ class CmsBootstrap {
         'core/modules/layout/layout.module',
       ),
       'Standalone' => array(
-        'data/civicrm.settings.php',
+        'civicrm.settings.config.php.standalone',
+        // or?
+        // 'data/civicrm.settings.php',
       )
     );
 
@@ -420,14 +422,17 @@ class CmsBootstrap {
 
       foreach ($cmsPatterns as $cmsType => $relPaths) {
         if (!empty($this->options['cmsType']) && $this->options['cmsType'] != $cmsType) {
+          $this->writeln("Skipping $cmsType");
           continue;
         }
         foreach ($relPaths as $relPath) {
           $matches = glob("$basePath/$relPath");
+          $this->writeln("Checking $basePath/$relPath");
           if (!empty($matches)) {
             return array('path' => $basePath, 'type' => $cmsType);
           }
           $matches = glob("$basePath/web/$relPath");
+          $this->writeln("Checking ($cmsType) $basePath/web/$relPath");
           if (!empty($matches)) {
             return array('path' => "$basePath/web", 'type' => $cmsType);
           }
--- a/src/CmsBootstrap.php
+++ b/src/CmsBootstrap.php
@@ -334,14 +334,27 @@ class CmsBootstrap {
    * @return $this
    */
   public function bootStandalone($cmsPath, $cmsUser) {
+    global $loggedInUserId;
     // Load the composer libs.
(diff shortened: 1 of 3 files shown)


src/CmsBootstrap.php
```diff
@@ -203,6 +217,25 @@ public function bootCivi() {
...
+        ->execute()->single();
+        \CRM_Core_Session::singleton()->set('ufId', $userID);
+        $loggedInUserId = $userID['contact_id'];
+      }
+      if (empty($loggedInUserId)) {
+        throw new \RuntimeException("Unable to login as '$this->deferredUserToLogin'");
+      }
+    }
+  }
+
```
artfulrobot, line 238: This depends on the standaloneusers extension, but I think it's ok for now.

src/Util/SetupCommandTrait.php
```diff
@@ -137,7 +137,14 @@ protected function bootSetupSubsystem(InputInterface $input, OutputInterface $ou
...
     if ($input->getOption('db')) {
-      $setup->getModel()->db = DbUtil::parseDsn($input->getOption('db'));
+      $model = $setup->getModel();
+      // Set db and also set cmsDb to the same db if standalone
+      $model->db = DbUtil::parseDsn($input->getOption('db'));
+
+      // todo: artfulrobot: we should only do this for standalone
+      if ('standalone fixme' && empty($model->cmsDb)) {
+        $model->cmsDb = $setup->getModel()->db;
+      }
```
artfulrobot, line 147: This is a fixme. Anyone help?

kainuk: Is the CMS database used somewhere by cv? If not you could skip this step.

kainuk:
I just tested this PR with the following actions

* cv flush
* cv sql:cli
* cv core:install (using a db string and a url).
* cv ext:download.

Currently, it is very useful addition for standalone administration.

src/Util/SetupCommandTrait.php
artfulrobot: If you skip this step you will get warnings from the installer about it being missing/credentials being wrong. The installer and boot process assumes to be able to boot the cms first, which means we have to oblige in various ways!

artfulrobot: But here ↑ we need code that detects whether the install type is Standalone, and only executes this `if {}` in that case.

src/CmsBootstrap.php
totten: Agree, that's OK for now. It sounds like it only activates on standalone deployments (which are still early-days).

totten:
civibot, add to whitelist

src/Util/SetupCommandTrait.php
totten: @artfulrobot I was hoping we could get these things merged to prevent MC's. It seemed the main issue was this `standalone fixme` conditional. After poking around, I think you're probably right that it's easier to oblige it in this way. I've pushed an update to use a real conditional.

As y'all indicate, there is a bit of impedance mismatch since the existence of a prior CMS was an original assumption of the `Civi\Setup` library -- and it's not the same for "Standalone". We may want to talk/shift more to account for that, but (for the moment) I think it's more useful to consolidate a bit so that we don't have as many patches floating around.

artfulrobot: @totten Thanks so much, Tim! Yeah, it seems there's quite a lot of pretend-we're-a-CMS which does involve a few hoops, but they're all quite big, low hoops that are fairly light to implement so far, just would look odd if you didn't know Civi's background.

totten:
Test fails are unrelated.

PR #143, opened 2023-02-07
Allow arguments to be passed to the script command 'cv scr'

aydun:
If `cv scr` is called with arguments like this:

`cv scr myscript.php arg1 arg2 ...`

`myscript.php` will have `$argv = ['arg1', 'arg2']`  

Quoted args are passed through correctly:

  `cv scr myscript.php one 'two and' three`  => `$argv = ['one', 'two and', 'three']`

Options starting with a hyphen can also be passed to the script if `--` is used:
  
  `cv scr myscript.php -- -a b`  => `$argv = ['-a', 'b']`

The superglobal `$GLOBAL['argv']` is also available with the full command line but `$argv` will usually be easier to work with.

### DIFF
--- a/src/Command/ScriptCommand.php
+++ b/src/Command/ScriptCommand.php
@@ -52,10 +52,10 @@ class ScriptCommand extends BaseCommand {
 
   /**
    * @param \Symfony\Component\Console\Output\OutputInterface $output
-   * @param $script
-   * @param array
+   * @param string $script
+   * @param array $scriptArguments
    */
-  protected function runScript(OutputInterface $output, $script, $scriptArguments) {
+  protected function runScript(OutputInterface $output, string $script, array $scriptArguments = []) {
     $output->writeln("<info>[ScriptCommand]</info> Start \"$script\"", OutputInterface::VERBOSITY_DEBUG);
     // This puts the script arguments in the same variable scope as the script
     // so scripts can access arguments using $argv $argc
--- a/src/Command/ScriptCommand.php
+++ b/src/Command/ScriptCommand.php
@@ -60,6 +60,7 @@ class ScriptCommand extends BaseCommand {
     // This puts the script arguments in the same variable scope as the script
     // so scripts can access arguments using $argv $argc
     $argv = $scriptArguments;
+    array_unshift($argv, $script);
     $argc = count($argv);
     require $script;
     $output->writeln("<info>[ScriptCommand]</info> Finish \"$script\"", OutputInterface::VERBOSITY_DEBUG);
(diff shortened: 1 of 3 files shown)


totten:
Oh, yes, this is very nice @aydun.  It even has a test! 😃 

There was one small catch -- the synthetic `$argv` was missing the the standard 0-th parameter. Compare:

```
$ php ~/src/cv/tests/Command/hello-args.php one two three
0: /Users/totten/src/cv/tests/Command/hello-args.php
1: one
2: two
3: three
```

```
$ cv scr ~/src/cv/tests/Command/hello-args.php one two three
0: one
1: two
2: three
```

I pushed up a revision to make it match. Unless there's an objection to the change, let's merge.

aydun:
Thanks @totten.  I realised that if the required script uses a variable called `$output`, it will overwrite the existing one resulting in errors.  This last change prevents that.

totten:
> (even if the first thing I'mma do whenever I use this is array_shift($argv)!).

Some rituals must be preserved.

>  I realised that if the required script uses a variable called $output ... 

Agree, this looks like better isolation now. 👍

aydun:
Thanks!

totten:
> (@artfulrobot) (even if the first thing I'mma do whenever I use this is array_shift($argv)!).

This is a bit of tangential rant, but it might be interesting... `argv` was one of my pet-peeves in scripting. Specifically, the need for impromptu parsing of argv. It doesn't really come up on tiny scripts (*eg 20 SLOC; not enough options to worry about parsing them*) or full programs (*eg 5000+ SLOC; big enough to setup a standalone project and download nice CLI library*). But for a certain size in the middle (50-500 SLOC), it's kind of annoying. I don't care to remember how many times I had to write an argv parse-loop for a script in this size-bracket.
 
Now-a-days, for scripts in that middle-space, I like to pull in "[silly](https://github.com/mnapoli/silly/)"/"[clippy](https://github.com/clippy-php/std#readme)". If you understand the synopsis from a typical Unix manpage, then the notation may be familiar. Ex:

```bash
## Goal:
myscript [--foo] [-b|--bar=] files*
```
```php
## Implementation:
$c['app']->main('[--foo] [-b|--bar=] files*', function($io, bool $foo, ?string $bar, array $files) {
  $io->writeln("The foo is " . 
    ($foo ? '<info>ON</info>' : '<error>OFF</error>'));
  $io->writeln("The value of bar is <info>$bar</info>");
  $io->writeln("The files are: " . json_encode($files));
});
```

In that example, the script accepts  `--foo`, `-b=123`, `-b 123`, `-b123`, `--bar 123`, `--bar=123`, and then an open-ended array of filenames. It's definitely nicer than`getopt` or an impromptu parse-loop.

Of course, I've only used it for plumbing that's outside of Civi's runtime... it could be that (in this context) the path-of-least-resistence is still writing an argv parser in-situ...

PR #144, opened 2023-02-16
document boolean params in JSON-ish Key-Value format

highfalutin:
see https://github.com/civicrm/civicrm-core/pull/25595


totten:
This is a good idea.

In the future (circa 5.60?), we might want to change the example to `useTrash` instead of `checkPermissions`. Strictly speaking,  `checkPermissions=0` isn't necessary (*because it matches the default*). But `useTrash` won't work for a while, so `checkPermissions` is probably the best demonstration.

PR #145, opened 2023-03-08
Bootstrap - Switch default boot process. Update docs.

totten:
`cv` includes supports two protocols for booting a full environment with CiviCRM+CMS.

| CLI Flag | Implementation | Description | Customization |
| -- | -- | -- | -- |
| `--level=full` | `Bootstrap.php` |  Original protocol. Adapted from `civicrm-core`'s`civicrm.config.php`. Boots Civi then CMS. | Configure `CIVICRM_SETTINGS`  |
| `--level=cms-full` | `CmsBootstrap.php` | Newer protocol. Adapted from typical HTTP lifecycle. Boots CMS then Civi. |  Configure `CIVICRM_BOOT` |

With this PR, `cv` will prefer `cms-full` (`CmsBootstrap.php`). It aims to phase-out the older `full` (`Bootstrap.php`) protocol. However, this is a notable policy change, and `CmsBootstrap` has not been tested as widely. The PR retains some options so that the site-builder can switch between them more easily.

Comments
-----

* The updated `README.md` describes more about how to use the options.
* Every now and then, people run into issues like https://lab.civicrm.org/dev/core/-/issues/4128 -- where a subtle discrepancy between the HTTP and CLI boot causes confusion.  Switching the default should reduce that kind of issue.
* `CmsBootstrap` ought to be better, but it hasn't been used much. And that won't change unless we make it the default. So we'll go ahead with the change. But if anyone needs a way to back out, then set:
    ```bash
    export CIVICRM_SETTINGS=Auto
    ```

Technical Details
---------

* Formally, most `cv` subcommands default to `--level=full|cms-full`. The PR changes the _interpretation_ of `full|cms-full`.
* Practically, for most invocations by most users, this changes the effective default from __`--level=full`__ to __`--level=cms-full`__. 
* The PR builds on a few ideas that were already in there:
    * `CIVICRM_BOOT` already provided a way to configure the `CmsBootstrap.php` protocol.
    * `CIVICRM_BOOT` already signaled an intent to activate `CmsBootstrap.php`.
    * `CIVICRM_SETTINGS` already provided a way to configure/ the `Bootstrap.php` protocol.
    * Now, `CIVICRM_SETTINGS` also signals an intent to activate `Bootstrap.php`.
* The `CIVICRM_BOOT` and `CIVICRM_SETTINGS` accept a few more values than before:
    * `CIVICRM_BOOT=Auto://.` will activate `CmsBootstrap.php` and allow it to search for the CMS.
    * `CIVICRM_SETTINGS=Auto` will activate `Bootstrap.php` and allow it to search for CiviCRM.


totten:
OK, there were a lot of failures relating to the `php:boot` command - because it didn't have a code-generator for `--level=cms-full`.  This is the command that works like so:

```php
eval(`cv php:boot`);
echo CRM_Utils_System::version();
```

So I added an implementation for `--level=cms-full`. The output looks fairly different from `--level=full`. Partially, that's because it takes a different approach to class-loading. (*Past-Me would have been concerned about this approach, but I think it might be OK since `cv.phar` now uses namespace prefixing.*)

It does point to another discrepancy though -- the `Bootstrap` populates `$_CV` but `CmsBootstrap` doesn't.

```
$ cv ev 'return $GLOBALS["_CV"]["ADMIN_USER"];' --level=full
"admin"
$ cv ev 'return $GLOBALS["_CV"]["ADMIN_USER"];' --level=cms-full
null
```

That will be a blocker - it's needed for various test-suites to work properly.

totten:
A second issue with switching default - on Backdrop, `cv http --login --user=admin ...` works with `--level=full` but fails with `--level=cms-full`.

PR #146, opened 2023-03-08
Fix `php:eval` exit-codes. Fix multiple test issues.

totten:
This fixes multiple issues in the test-suite. Most of the commits are small the things.

The gnarly part involves "EvalCommand", which had both direct failures (`EvalCommandTest`) and various indirect failures (`ExtensionLifecycleTest`). It seems likely that this was a leftover from the last Symfony upgrade.

Intuitively: if `cv php:eval` encounters an error, then the exit-code should indicate as much. But in some cases, it didn't. This is because of how the various error-handlers (Symfony, Drupal, etc) interacted with each other. Structurally, I'm not keen on the patch -- but functionally, it's what we need, and I can't think of anything else to do the needful. It passes the tests on D7+WP, php73+php81. So it's certainly better.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,9 @@
         "platform": {
             "php": "7.2.0"
         },
-        "bin-dir": "bin"
+        "bin-dir": "bin",
+        "allow-plugins": {
+            "civicrm/composer-downloads-plugin": true
+        }
     }
 }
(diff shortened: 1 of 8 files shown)


totten:
civibot, test this please

totten:
There's still a couple more test issues affecting D9+WP, but they're pre-existing. Can do those as separate PRs.

Issue #148, opened 2023-03-08
in drupal8+, calling `cv --user=<user> flush` before the user has logged in results in backtrace


jmcclelland:
If I install:

 * Drupal8+ via drush (specifying a user "admin" and password on the command line)
 * civicrm via cv

And then:
 * Try to execute the `cv --user=admin flush` I get:

```
www-data@b2378ad6fbcf:~/powerbase/web$ cv --user=admin flush                                                                                                                                                         
Error: Call to undefined method Drupal\Core\Session\AccountProxy::get() in /var/www/powerbase/vendor/civic                                                          
rm/civicrm-core/CRM/Utils/System/Drupal8.php on line 596 #0 /var/www/powerbase/vendor/civicrm/civicrm-core                                              
/CRM/Utils/System/Base.php(893): CRM_Utils_System_Drupal8->getUserIDFromUserObject()                                                           
#1 /var/www/powerbase/vendor/civicrm/civicrm-core/CRM/Core/BAO/UFMatch.php(71): CRM_Utils_System_Base->get                               
BestUFID()                                                                                                                                            
#2 phar:///usr/local/bin/cv/src/Util/BootTrait.php(207): CRM_Core_BAO_UFMatch::synchronize()                                                                                   
#3 phar:///usr/local/bin/cv/src/Util/BootTrait.php(112): Civi\Cv\Command\FlushCommand->ensureUserContact()               
#4 [internal function]: Civi\Cv\Command\FlushCommand->_boot_full()                                                  
#5 phar:///usr/local/bin/cv/src/Util/BootTrait.php(45): call_user_func()                                      
#6 phar:///usr/local/bin/cv/src/Command/FlushCommand.php(29): Civi\Cv\Command\FlushCommand->boot()                                                                         
#7 phar:///usr/local/bin/cv/vendor/symfony/console/Command/Command.php(255): Civi\Cv\Command\FlushCommand-                                     
>execute()                                                                                                                                       
#8 phar:///usr/local/bin/cv/vendor/symfony/console/Application.php(1009): Symfony\Component\Console\Comman                                                             
d\Command->run()                                                                                                                                                               
#9 phar:///usr/local/bin/cv/vendor/symfony/console/Application.php(273): Symfony\Component\Console\Applica                                     
tion->doRunCommand()                                                                                                                     
#10 phar:///usr/local/bin/cv/src/Application.php(82): Symfony\Component\Console\Application->doRun()                                                  
[...]

PR #149, opened 2023-03-08
Drupal 9 - Fix error regarding `AccountProxy::get()` (#148)

totten:
This should fix an issue that appears in both #148 and in the test-matrix. It presents as:

```
Error: Call to undefined method Drupal\Core\Session\AccountProxy::get() in
CRM_Utils_System_Drupal8->getUserIDFromUserObject() (line 595 of
/home/jenkins/bknix-max/build/build-1/vendor/civicrm/civicrm-core/CRM/Utils/System/Drupal8.php).
```

And it flows from a call to `\CRM_Core_BAO_UFMatch::synchronize`. There are, in fact, two callers for `synchronize()`, and I believe both need updating.

### DIFF
--- a/src/CmsBootstrap.php
+++ b/src/CmsBootstrap.php
@@ -456,7 +456,8 @@ class CmsBootstrap {
         break;
 
       case 'Drupal8':
-        \CRM_Core_BAO_UFMatch::synchronize(\Drupal::currentUser(), TRUE, CIVICRM_UF, 'Individual');
+        $user = \Drupal\user\Entity\User::load(\Drupal::currentUser()->id());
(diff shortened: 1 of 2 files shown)


totten:
From a test POV, here's the [before](https://test.civicrm.org/job/CiviCRM-Cv-PR/7/) and [after](https://test.civicrm.org/job/CiviCRM-Cv-PR/8/). So that `drupal-9` column looks better.

I'll hold off merging for a moment in case @jmcclelland or @seamuslee001 has any other thoughts.

seamuslee001:
This looks right to me and matches what we do in the CiviCRM drupal 8 module's controller as well

PR #150, opened 2023-03-09
Allow explicit-automatic flags

totten:
(This is extracted from the bigger #145, which needs some additional work before reaching its policy goal of flipping the defaults.)

Both `Bootstrap.php` and `CmsBootstrap.php` can be customized by environment variables (`CIVICRM_SETTINGS` and `CIVICRM_BOOT`, respectively). This patch adds a couple more options:

__Before__: The values must concretely define the exact location of the `civicrm.settings.php` or of the CMS (respectively).

* For `Bootstrap`, set `CIVICRM_SETTINGS=/var/www/sites/default/civicrm.settings.php`
* For `CmsBootstrap`, set `CIVICRM_BOOT=Drupal://var/www`

__After__: The old values still work. Additionally, you can specifically indicate a preference for _the default/automatic search behavior_, i.e.

* For `Bootstrap`, set `CIVICRM_SETTINGS=Auto` (literally, "automatically find the settings file")
* For `CmsBootstrap`, set `CIVICRM_BOOT=Auto://.` (literally, "automatically choose the CMS based on the current folder")

The two notations are slightly because the two variables are already different. They aim to feel "ok" when compared to their other regular values.

### DIFF
--- a/src/Bootstrap.php
+++ b/src/Bootstrap.php
@@ -61,7 +61,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  *     type of CMS is being used.
  *     (Default: NULL)
  *   - env: string|NULL. The environment variable which may contain the path to
- *     civicrm.settings.php. Set NULL to disable.
+ *     civicrm.settings.php (or the token "Auto"). Set NULL to disable environment-checking.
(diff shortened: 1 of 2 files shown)


totten:
Test failures appear unrelated/pre-existing.

PR #151, opened 2023-03-09
SetupCommandTrait - Autoload extra plugins for "Standalone" installation

totten:
(ping @artfulrobot)

## Before

When running `cv core:install`, it can load `*.civi-setup.php` plugins from:

* (1) `{$civicrmRoot}/setup/plugins`
* (2) A list of supplemental folders, specified by `--plugin-path=/my/extra/plugins`.

## After

It will also load plugins from:

* (3) `{$cmsRoot}/setup/plugins` (if the CMS is `Standalone`)

## Technical Details

This also changes the way in which it scans the extra plugin paths -- it will look one level deeper. This means that the file-tree under `{$cmsRoot}/setup/plugins/` can look like the file-tree under `{$civicrmRoot}/setup/plugins`. Moreover, since the file-trees match, you can also do an override (*although I wouldn't recommend doing that for more than a few weeks*).

### DIFF
--- a/src/Util/SetupCommandTrait.php
+++ b/src/Util/SetupCommandTrait.php
@@ -1,6 +1,7 @@
 <?php
 namespace Civi\Cv\Util;
 
+use Civi\Cv\CmsBootstrap;
 use Civi\Setup\DbUtil;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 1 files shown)


totten:
civibot, test this please

totten:
New test run looks better.

@artfulrobot Yup, exactly. And I did some local testing last night. It was along the lines of:

```bash
cd build/standalone
mkdir -p setup/plugins/init
mv vendor/civicrm/civicrm-core/setup/plugins/init/Standalone.civi-setup.php setup/plugins/init/
civibuild reinstall 
```

And that worked.

PR #152, opened 2023-03-10
Let admin choose default boot proto. Respect both CIVIRM_BOOT and CIVCRM_SETTINGS.

totten:
(*Extracted from #145.*)

This is preparation for `v0.4.x`. The ultimate aim is to flip the default boot behavior (`full/Bootstrap.php` <=> `cms-full/CmsBootstrap.php`); but that is a big change. This patch allows admins fine-tune the transitional period.

* In v0.3, the default is `full`/`Bootstrap.php`.
* In v0.4, it will be `cms-full`/`CmsBootstrap.php`.
* If you want to get ahead (*start using `CmsBootstrap.php`  by default; now*), then you can explicitly set `CIVICRM_BOOT=Auto://`. This setting will be respected on either version.
* If you want to stay behind (*continue using `Bootstrap.php` by default*), then you can explicitly set `CIVICRM_SETTING=Auto`. This setting will be respected on either version.

These env-vars already exist, but they're not consistently respected.  You sometimes need to manually set `--level=full` or `--level=cms-full`.  This patch makes it easier to switch -- you can express a preference for either style, and you don't need to manually pass `--level=X` with each call.

This PR also updates the README to describe the options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,15 +6,9 @@ The `cv` command is a utility for interacting with a CiviCRM installation. It pe
 Requirements
 ============
 
-A local CiviCRM installation.
-
-PHP v5.6+.
-
-Support may vary depending on the host environment (CMS type, file-structure, symlinks, etc).
- * *Tested heavily*: Drupal 7/8 single-site, WordPress single-site, UnitTests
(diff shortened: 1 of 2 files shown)


totten:
Test failures are unrelated/pre-existing.

PR #153, opened 2023-03-10
Remove unused functions. Simplify dependencies.

totten:
* __Specifically__, it's getting harder to re-embed the `Bootstrap.php` and `CmsBootstrap.php` because of dependencies. So remove some.
* __Generally__, `symfony/*` can be an annoyance for upgrades. If we can remove one that isn't strongly needed, then great.


totten:
Test failures are pre-existing/unrelated.

PR #154, opened 2023-03-10
Improve interoperability of `Bootstrap.php`/`CmsBootstrap.php`

totten:
(Extracted from #145) 

As part of the effort to switch from `Bootstrap.php` to `CmsBootstrap.php`, we need to make sure that it actually works as replacement. This addresses a couple edge-cases:

1. The `php:boot` command should support `--level=cms-full`
2. When using `--level=cms-full`, you should get the same `$_CV` variables (as currently provided by `--level=full`).


totten:
Test failures unrelated.

Issue #147, opened 2023-03-08
Error: Call to undefined method Symfony\Component\Console\Helper\Helper::width()


calbasi:
My install: Drupal 10, CiviCRM 5.60alpha1, php8.1

When executing a test command, like:
`/usr/bin/php8.1 vendor/bin/cv ext:list`

I get this error:

> Using extension feed "https://civicrm.org/extdir/ver=5.60.alpha1|uf=Drupal8|status=stable|ready=ready"
> Error: Call to undefined method Symfony\Component\Console\Helper\Helper::width() in /var/www/clients/client4/web5/web/vendor/symfony/console/Helper/Table.php on line 834 #0 /var/www/clients/client4/web5/web/vendor/symfony/console/Helper/Table.php(815): Symfony\Component\Console\Helper\Table->getCellWidth()
> #1 /var/www/clients/client4/web5/web/vendor/symfony/console/Helper/Table.php(381): Symfony\Component\Console\Helper\Table->calculateColumnsWidth()
> #2 phar:///var/www/clients/client4/web5/web/vendor/bin/cv/src/Util/StructuredOutputTrait.php(191): Symfony\Component\Console\Helper\Table->render()
> #3 phar:///var/www/clients/client4/web5/web/vendor/bin/cv/src/Command/ExtensionListCommand.php(84): Civi\Cv\Command\ExtensionListCommand->sendTable()
> #4 phar:///var/www/clients/client4/web5/web/vendor/bin/cv/vendor/symfony/console/Command/Command.php(255): Civi\Cv\Command\ExtensionListCommand->execute()
> #5 phar:///var/www/clients/client4/web5/web/vendor/bin/cv/vendor/symfony/console/Application.php(1009): Symfony\Component\Console\Command\Command->run()
> #6 phar:///var/www/clients/client4/web5/web/vendor/bin/cv/vendor/symfony/console/Application.php(273): Symfony\Component\Console\Application->doRunCommand()
> #7 phar:///var/www/clients/client4/web5/web/vendor/bin/cv/src/Application.php(82): Symfony\Component\Console\Application->doRun()
> #8 phar:///var/www/clients/client4/web5/web/vendor/bin/cv/vendor/symfony/console/Application.php(149): Civi\Cv\Application->doRun()
> #9 phar:///var/www/clients/client4/web5/web/vendor/bin/cv/src/Application.php(49): Symfony\Component\Console\Application->run()
[...]

PR #158, opened 2023-03-13
php-scoper - Fix prefixing

totten:
Per #147, it appears that the prefixing is not currently working for `cv.phar`. This should restore prefixing.

Before
---------

* Rely on `patchers` to enable/disable prefixing for specific symbols
* Nothing will complain about incorrect prefixing
* Compile on box 3.16.0 and php74

After
-------

* Rely on `exclude-classes`, `exclude-namespaces`, etc to enable/disable prefixing for specific symbols
* Build script will spot-check some prefixes and complain if they're wrong.
* Compile on box 4.3.7 and php81

Technical Details
----------------------

I played around with the `scoper.inc.php` rules on 3.16.0, and the problem seemed to be caused by the `patchers` function. This is somewhat fiddly way to enable/disable prefixing, and it's no longer necessary since `exclude-classes` and `exclude-namespaces` can do the needful.

After addressing that, there was another issue where `polyfill-php80` didn't load correctly. This required upgrading `box` and `php`.

I tested the resulting `cv.phar` lightly on `php73` with D7, WP, and D9. They all seemed OK.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -12,7 +12,7 @@ PRJDIR=$(absdirname "$0")
 OUTFILE="$PRJDIR/bin/cv.phar"
 set -e
 
-BOX_VERSION=3.16.0
+BOX_VERSION=4.3.7
 BOX_URL="https://github.com/box-project/box/releases/download/${BOX_VERSION}/box.phar"
 BOX_DIR="$PRJDIR/extern/box-$BOX_VERSION"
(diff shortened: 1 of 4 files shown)


totten:
There was an issue when using `--level=cms-first` or `--level=cms-full` (as in `core:install`). Updated prefixing rules with b91a3e7f79d147bc75366b2dc0f78aa4f8361512.

PR #160, opened 2023-05-09
'cv path' - Fix warning for top-level lookups on PHP 8.1

totten:
`cv path` is often used to locate a file, e.g.

* `cv path -x my_ext/foo/bar`
* `cv path -c extensionsDir/foo/bar`

But sometimes it's used without a specific file, e.g.

* `cv path -x my_ext`
* `cv path -c extensionsDir`

The latter can emit a warning PHP 8.1. This patch fix it.


totten:
Test failures unrelated. (They're reporting that routing for CiviCRM 5.57+5.51 are not working in Drupal 9.5.9, which is a separate/pre-existing problem being worked on separately.)

Issue #164, opened 2023-06-30
Shell completion


chriscroome:
Is there a shell completion (I'm specifically interested in a Bash one) available for `cv` anywhere? 

If not would adding a method to generate the required files be something that could be considered as a feature request for the future?

totten:
@chriscroome Shell completion would be great. It doesn't exist - mostly because I don't know the protocol. However, [this library](https://github.com/stecman/symfony-console-completion) seems to provide some support (without needing to learn the protocol..?).

I'm not sure if I have time to do flesh it out, but here's a quick/experimental stab at #165. If that seems OK, then it's patch-welcome for further tuning.

chriscroome:
Sorry I didn't look at this sooner @totten I have now tested it and it's great :slightly_smiling_face: I've added installing it to a [cv Ansible role I've written](https://galaxy.ansible.com/chriscroome/cv), thanks!

PR #165, opened 2023-06-30
Basic autocomplete support (#164)

totten:
Quick/experimental support for bash autocomplete:

[asciicast](https://asciinema.org/a/b7jVdCXqaz18tx2W4MemvL0jW)

Possible issues:

* Needs to spin up `cv` every time you press `Tab`. That might be a bit heavy? The alternative would be a smaller/dedicated script?
* It uses the metadata from Symfony Console -- so it can complete some symbols (e.g. subcommand names and some `--foo` options). But completing other things (such as entity-names, action-names, or extension-keys) would require some more work.


totten:
Test failure is unrelated/pre-existing. (*The current point-release of D9 has a regression that affects Civi E2E testing.*)

totten:
civibot, test this please

totten:
Test failure unrelated.

PR #168, opened 2023-07-12
Define subproject `cv-lib` for use by `civix` (et al).

totten:
Overview
------------

Sometimes it's handy to use the `CmsBootstrap` or `Bootstrap` from another program (such as `civix` or a throw-away pogo script).

Before
---------

Options for downstream tools are either:

1. Depend on `civicrm/cv`. This will pull in the necessary library -- and a lot of other stuff (`symfony/console`, `symfony/process`, `psysh`, etc). This makes the dependency a bit onerous.
2. Download individual PHP files (eg `Bootstrap.php`). However, this is also hard to manage as `Bootstrap.php` evolves -- either the downstream dev needs to track the list of class-dependencies, or the upstream dev needs to avoid introducing any kind of class-dependencies.

After
------

Downstream tools can do `composer require civicrm/cv-lib`.

Comments: OutputInterface vs LoggerInterface
--------------

`Bootstrap` and `CmsBootstrap`  generate some log/debug messages. This has traditionally been displayed via `$options['output']` (an instance of `OutputInterface` per `symfony/console`). 

Although the handling of `$options['output']` was somewhat guarded, it still amounted to a hard dependency on `symfony/console` -- which would make it awkward to include in other apps.

So I reorganized as two options:

* `$options['log']` (`psr/log`'s `LoggerInterface`)
* `$options['output']` (`symfony/console`'s `OutputInterface`)

The overall effect: `LoggerInterface` is now the main way it emit log messages. Support for `OutputInterface` is still there, but it's entirely optional, and it will be adapted to fit `LoggerInterface`.

Comments: Git and Composer
--------------

I want to keep `civicrm/cv` as the CLI project -- because some site-builders have added that to their top-level `composer.json`. And I also want to keep `civicrm/cv` as unified repo (*because it's easier to QA that way*). But `packagist` requires publishing `civicrm/cv-lib` as a distinct git repo.

So that's a little weird: `cv.git` has the code for both `cv` (*app*) and `cv` (*library*). Then `cv-lib.git` has another copy of the library-code. However, composer [defines an option for this situation](https://getcomposer.org/doc/04-schema.md#replace). You can see an example in how [`symfony/symfony`  replaces all of its subprojects (`symfony/yaml`, etc)](https://github.com/symfony/symfony/blob/6.4/composer.json#L59-L68).

Comments: Order of operations
----------------------

I believe the order of operations should be:

1. Reorganize `cv.git`. Open PR. Ensure that it's passing test.
2. Publish first-draft of `cv-lib.git`. Try using `cv-lib` in a downstream project.
3. Merge PR on `cv.git`.
4. Update and tag `cv.git` and `cv-lib.git`.
5. Again, try using `cv-lib.git` in a downstream project.
6. Sort out release-process pushing future updates to `cv-lib.git`.


totten:
Tests passed.

I did a trial-run -- pushing `cv-lib` to a public repo (https://github.com/civicrm/cv-lib) and then using that with `civix` (both `~/src/civix/bin/civix` and `civix.phar`) on a few different UFs (D7, D9, WP). Seems promising enough.

totten:
Note to self: For first release, I basically used this script:

```bash
#!/usr/bin/env bash
set -e

## Experimental!

GUARD=echo
BRANCH="master"
TAG="$1"
LIB_AUX_DIR=/tmp/cv-lib
LIB_AUX_URL='git@github.com:civicrm/cv-lib.git'

if [ -z "$TAG" ]; then
  echo >&2 "usage: tag.sh <version>"
  exit 1
fi

if [ -d "$LIB_AUX_DIR" ]; then
  $GUARD rm -rf "$LIB_AUX_DIR"
fi

$GUARD git checkout "$BRANCH"
$GUARD git subrepo push lib

$GUARD git tag "$TAG"
$GUARD git push origin "$BRANCH" "$TAG"

$GUARD git clone -b "$BRANCH" --depth 1 "$LIB_AUX_URL" "$LIB_AUX_DIR"
$GUARD pushd "$LIB_AUX_DIR"
$GUARD git tag "$TAG"
$GUARD git push origin "$TAG"
$GUARD popd
```

Also: The packagist.org registration is up. The webhook *should* be configured, but we won't know until there's another update.

PR #169, opened 2023-07-13
Logging - Don't require psr/log for `cv-lib`. Fix STDERR/STDOUT routing.

totten:
This is a multi-prong update to logging. There are two major themes:

# Don't require psr/log

In [discussion about civix using cv-lib](https://github.com/totten/civix/pull/300#issuecomment-1632399428), we found that some environments have contention over the substance of `psr/log`. (`cv-lib` may load its copy of `psr/log` before the UF starts, and UF's may have different expectations for the version of `psr/log`.) And the reality is that `psr/log` has several major versions in the wild.

So as a defensive move, it's better if `cv-lib` doesn't have any specific expectations about `psr/log`.

# Cleanup STDERR/STDOUT distinction

When you enable verbose output (`cv -vvv`), `cv` has been a bit confused about which information should go to STDOUT and which to STDERR.

Since `cv` (and `cv-lib`) are intended to be used in many kinds of scripts (*including scripts with structured output - JSON, SQL, PHP, etc*), it's bad default to send any logs to STDOUT. The verbose logs should go to STDERR.

SO there are a few patches to make that change.


totten:
Most tests are passing. The failure appears to be random network I/O issues (e.g. `Could not resolve host: github.com`). When I re-ran, it just flip-flopped somethings with (different) random network I/O issues. So... I don't think the test-suite is finding any problems in the code.

demeritcowboy:
This works for me to resolve the drupal 10 problem in https://github.com/totten/civix/pull/300. Thanks @totten.

PR #176, opened 2023-08-07
(Future work) Define statement-level helper (`Top`) to bypass prefixing

totten:
This is a draft towards simplify `scoper.inc.php`. I don't need the simplification at quite this minute, and it would need some more work/testing. But I want to keep the patch around as a reminder for how to do it (*the next that `scoper.inc.php` comes to mind).

## Before

* `Bootstrap.php` and `CmsBootstrap.php` need to reference various UF symbols (`drupal_foobar`,  `JFactory`, `wp_*`, etc.)
* `scoper.inc.php` has a bunch of exclusions so that namespace-prefixes are not applied to these symbols.
* The exclusions are sync'd up between `cv.git` and `civix.git`.

## After

* `Bootstrap.php` and `CmsBootstrap.php` reference various UF symbols ( `Top::call('drupal_foobar')`, `Top::call('JFactory::getUser')`, etc)
* `scoper.inc.php` don't need those exclusions.
* Since he exclusions aren't needed, they don't need to be sync'd.

## TODO

Patch currently defines the `Top` helper. Next: Swap the various calls in `Bootstrap.php` and `CmsBootstrap.php` to use `Top`. Remove the rules. Re-tests both cv and civix.

### DIFF
--- a/lib/src/Top.php
+++ b/lib/src/Top.php
@@ -1,0 +1,93 @@
+<?php
+
+namespace Civi\Cv;
+
+class Top {
+
+  /**
+   * @var string|null
+   */
+  private static $prefixRegex;
+
+  /**
+   * Call a top-level function or method
+   *
+   * @param string|string[] $callable
+   *   Ex: 'drupal_bootstrap'
+   *   Ex: '\drupal_bootstrap'
+   *   Ex: ['Drupal', 'service']
+   *   Note: All symbols are implicitly relative to top-level namespace.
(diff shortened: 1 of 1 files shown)


totten:
This PR was left in "Draft" because it had the ambition to convert more things. But that's not really a blocker.

I found the same patch useful as part of another branch (*with a tweak*). I'm just gonna close this PR and then include the revised commit as part of a bigger PR.

PR #178, opened 2023-09-08
refs #composer updated Symfony components

kewljuice:


totten:
civibot, add to whitelist

composer.json
```diff
@@ -9,14 +9,14 @@
         }
     ],
     "require": {
-        "php": ">=7.3.0",
+        "php": ">=8.0.2",
```
totten, line 12: @kewljuice Yeah, that's not gonna pass - since `cv` and `cv.phar` need to be runnable on all supported versions of CiviCRM.

> For my installation with Drupal 10, I need the ^6 versions of the symfony libraries, the "symfony-console-completion" lacks this dependency. I read that stecman/symfony-console-completion:0.12 no longer supports older versions of symfony, only 6.3.

That is a conundrum. Some ways to go at that:

* It sounds like you're adding the `cv` source-code to the D10 build.  I would strongly suggest trying `cv.phar`.
    * `cv.phar` benefits from "namespace-prefixing" -- which allows `cv.phar` to use a singular version of each library (eg `symfony/console:^4`) regardless of what's happening with different CMSs (D7 vs D9 vs D10).
    * For a while (*last year?*), `cv.phar`s namespace prefixing was broken. This was fixed ~6 months ago (esp #158) and has direct+indirect test-coverage. The test+demo systems (eg https://d10-master.demo.civicrm.org) rely on`cv.phar`.
    * If you want `cv` embedded in the site/project, then consider [phive or composer-downloads-plugin](https://github.com/civicrm/cv/blob/master/doc/download.md#phar-composer) -- these options allow a workflow that's similar to `composer install` but retain support for namespace-prefixing.
* From my POV, `stecman/symfony-console-completion` is kind of experimental -- not essential.
    * One could probably make it a soft dependency -- i.e. `suggest` instead of `require`; for PHP logic, use some guard like `if (class_exists(...))`.
    * We would want to make sure that the `cv.phar` builder gets the `suggest`ed packages. (*Not certain how to do that off the top of my head, but it's probably possible...*)

totten:
* For PHP 8.4 compat, there have been a few PRs which updated Symfony components from 4=>5.
* So I still think the easiest way to deploy on D8/9/10/11-style composer-based websites is to pull in the PHAR. In the intervening time since the comments above, we added a slightly nicer way to do this:
    ```
    composer require civicrm/cli-tools
    ```
* As long as civicrm-core supports PHP 7, `cv` will (basically) have to continue treating Symfony 4 as canonical.
* But, in theory, it may valid to relax the constraints to allow `~5.4|~6.0` (or somesuch) and update `cv.git:src/*php` to be dual compatible. It's not a huge priority for me, but it's worked in some projects, and patches in that direction would be merge-able.

PR #179, opened 2023-09-13
use correct types in params to file_get_contents

highfalutin:
in response to a deprecation notice


colemanw:
@civicrm-builder add to whitelist

PR #184, opened 2023-11-16
cv api4 - Add support for more `where` operators

totten:
# Before:

The `where` shorthand (`+w`) supports some operators, like `LIKE`:

```bash
cv api4 Entity.get -T +w 'name LIKE Opti%' +s name
```

# After:

The `where` shorthand (`+w`) supports more operators, like `CONTAINS` or `REGEXP`:

```bash
cv api4 Entity.get -T +w 'type CONTAINS ManagedEntity' +s name

cv api4 Entity.get -T +w 'name NOT REGEXP .*Field' +s name
```

### DIFF
--- a/src/Util/Api4ArgParser.php
+++ b/src/Util/Api4ArgParser.php
@@ -5,6 +5,15 @@ namespace Civi\Cv\Util;
 class Api4ArgParser extends AbstractPlusParser {
 
   /**
+   * @var string
+   */
+  private $operators;
+
+  public function __construct() {
+    $this->operators = '\<=|\>=|=|!=|\<|\>|IS NULL|IS NOT NULL|IS EMPTY|IS NOT EMPTY|LIKE|NOT LIKE|IN|NOT IN|CONTAINS|NOT CONTAINS|REGEXP|NOT REGEXP|REGEXP BINARY|NOT REGEXP BINARY';
+  }
+
+  /**
    * @param array $params
    * @param string $type
    * @param string $expr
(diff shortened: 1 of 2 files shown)


totten:
Test was passing, except for some odd failures in [5.51,wp-demo,min](https://test.civicrm.org/job/CiviCRM-Cv-PR/71/BKPROF=min,BLDTYPE=wp-demo,CIVIVER=5.51,SUITE=cv-std,label=bknix-tmp/). Probably unrelated, but will fire one more run.

totten:
civibot, test this please

colemanw:
@totten the full list of accepted operators can be obtained by calling `\Civi\Api4\Utils\CoreUtil::getOperators()`

totten:
Thanks, good point @colemanw. Updated to prefer a live list.

src/Util/Api4ArgParser.php
```diff
@@ -4,6 +4,23 @@
...
+
+  public function __construct() {
+    if (is_callable(['\Civi\Api4\Utils\CoreUtil', 'getOperators'])) {
+      // 5.30+
+      $this->operators = implode('|', ArrayUtil::map(\Civi\Api4\Utils\CoreUtil::getOperators(), function($op) {
+        return preg_quote($op, '/');
+      }));
+    }
+    else {
+      $this->operators = '\<=|\>=|=|!=|\<|\>|IS NULL|IS NOT NULL|IS EMPTY|IS NOT EMPTY|LIKE|NOT LIKE|IN|NOT IN|CONTAINS|NOT CONTAINS|REGEXP|NOT REGEXP|REGEXP BINARY|NOT REGEXP BINARY';
```
colemanw, line 20: I mean, if you really wanna support < 5.30, then this list is a bit long. Those weren't all available at the time.

totten: Well... *shrug*. I just don't want the parser to crash where it was non-crashing before.

As to which operators work on which exact versions -- that's not really the concern of `cv`. The goal is to pass-through data to `civicrm_api4()` -- then `civicrm_api4()` can figure out whether it likes the request. (*Same as with entity-names and action-names.*)

PR #199, opened 2024-06-21
With pear/log 1.14+ the filename property no longer exists

demeritcowboy:
There is no way to get the filename anymore from the Log object. We instead made CRM_Core_Error::generateLogFileName() publicly callable.

I tested and this works with both earlier and new versions of pear/log.

I'm not sure where tests run but I expect ShowCommandTest will be failing after yesterday with an undefined property warning.


totten:
The test is happy. You're happy. I'm happy. The repo is happy. Let's merge!

PR #200, opened 2024-06-21
Combine "upgrade:db" and "ext:upgrade-db". Support "updb" alias.

totten:
Before
-------

* `cv updb` does not exist.
* `cv ext:upgrade-db` runs the extension upgrades.
* `cv upgrade:db` runs the core upgrade. Sometimes, it may also run extension upgrades:
    * If the core schema is up-to-date, then it does nothing. (Neither core nor extension.)
    * If the core schema is out-of-date, then it runs core upgrades. If the version is >=5.53, then it also runs extension upgrades.

After
-------

* `cv updb` is an alias for `cv upgrade:db`
* `cv ext:upgrade-db` is deprecated
* `cv upgrade:db` runs core and/or extension upgrades.
    * If the core schema is out-of-date, then it runs both core and extension upgrades. (This is also known as `--mode=full`.)
    * If the core schema is up-to-date, then it just runs the extension upgrades. (This is known as `--mode=ext`.)

Screenshot 2024-06-21 at 6 36 12 PM

Comments
-----------

The traditional command `upgrade:db` has been more featureful -- it provides progress-indicators, verbose output, and resume/skip options... but `ext:upgrade-db` had none of these. With the combined command, these features are available in both contexts.

### DIFF
--- a/src/Command/ExtensionUpgradeDbCommand.php
+++ b/src/Command/ExtensionUpgradeDbCommand.php
@@ -31,6 +31,7 @@ Note:
   }
 
   protected function execute(InputInterface $input, OutputInterface $output) {
+    $output->writeln("<error>WARNING: \"ext:upgrade-db\" is deprecated. Use the main \"updb\" command instead.</error>");
     $this->boot($input, $output);
 
     $output->writeln("<info>Applying database upgrades from extensions</info>");
--- a/src/Command/UpgradeDbCommand.php
+++ b/src/Command/UpgradeDbCommand.php
@@ -19,12 +19,14 @@ class UpgradeDbCommand extends BaseCommand {
   protected function configure() {
     $this
       ->setName('upgrade:db')
+      ->setAliases(['updb'])
       ->setDescription('Run the database upgrade')
       ->configureOutputOptions(['fallback' => 'pretty'])
       ->addOption('dry-run', NULL, InputOption::VALUE_NONE, 'Preview the list of upgrade tasks')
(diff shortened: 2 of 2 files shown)


colemanw:
This looks good, although I find the difference between "auto" and "full" to be a bit confusing. Both modes will upgrade everything, so is there really a need for the distinction?

totten:
Cool. Glad it's merged.

> ... find the difference between "auto" and "full" to be a bit confusing ...

Yeah, unfortunately, the pre-existing protocols for `CRM_Upgrade_Form` and `CRM_Extension_Upgrades` don't do what one wants for a unified `cv updb`. (*There's disagreement on whether to proceed with execution when the DB/code versions already match. `CRM_Upgrade_Form` protocol says to quit -- but `cv updb` should continue with extension-upgrades.*)

The mismatch means that some messiness is going to show up somehow. The `$mode` is just a way to organize that bit of mess. The options are distinguished as:

1. `auto` (*normal*): Do what a user actually expects from a unified DB upgrade command. (*Upgrade core and/or extension, depending on what's pending.*)
2. `full`: Follow the pre-existing protocol for `CRM_Upgrade_Form` (*as currently understood circa 5.74; with an extra compatibility step for <5.53*).
3. `ext`: Follow the pre-existing protocol for `CRM_Extension_Upgrades` (*as currently understood circa 5.74*)

Binding `$mode` to `--mode` (*with a default `auto`*) reflects an attitude of favor transparency. (*"If `$mode` has to exist, and if it makes a big difference in how the thing behaves, then `$mode` should probably be visible.*)

PR #211, opened 2024-09-06
why

demeritcowboy:
I'm curious about:
* Whether the automated tests will show anything.
* Why is this error handling needed.
  * I've previously had silent fails and needed to remove one or more of these to see the problem. Something similar recently came up in chat.

### DIFF
--- a/lib/src/Bootstrap.php
+++ b/lib/src/Bootstrap.php
@@ -213,7 +213,7 @@ class Bootstrap {
 
       $this->log->debug("Load settings file \"" . $settings . "\"");
       define('CIVICRM_SETTINGS_PATH', $settings);
-      $error = @include_once $settings;
+      $error = include_once $settings;
       if ($error == FALSE) {
         $this->log->notice("Failed to load settings file");
         throw new \Exception("Could not load the CiviCRM settings file: {$settings}");
--- a/lib/src/CmsBootstrap.php
+++ b/lib/src/CmsBootstrap.php
@@ -171,10 +171,6 @@ class CmsBootstrap {
     call_user_func([$this, $func],
       $cms['path'], $this->options['user'], $this->options['httpHost']);
 
-    if (PHP_SAPI === "cli") {
-      error_reporting(1);
-    }
-
     $this->log->debug("Finished");
     $this->bootedCms = $cms;
     return $this;
--- a/src/Application.php
+++ b/src/Application.php
@@ -43,9 +43,7 @@ class Application extends \Symfony\Component\Console\Application {
     $application = new Application('cv', static::version() ?? 'UNKNOWN');
 
     $application->setAutoExit(FALSE);
-    ErrorHandler::pushHandler();
     $result = $application->run();
-    ErrorHandler::popHandler();
     exit($result);
   }
 


lib/src/Bootstrap.php
```diff
@@ -213,7 +213,7 @@ public function boot($options = array()) {
 
       $this->log->debug("Load settings file \"" . $settings . "\"");
       define('CIVICRM_SETTINGS_PATH', $settings);
-      $error = @include_once $settings;
+      $error = include_once $settings;
```
totten, line 216: For this one, my guess: inertia

It's a little hard to see now, but `Bootstrap.php` arose from reorganizing/consolidating various copies of `civicrm.config.php.XXX` -- and some copies [had suppression on this line](https://github.com/civicrm/civicrm-drupal/blob/7.x-4.3/civicrm.config.php.drupal#L106). My guess is that someone felt dissatisfied that `include_once` didn't provide enough recoverability, and no one had the mix of incentive/time/expertise to question it.

But I share your view that this line probably isn't doing much.

lib/src/CmsBootstrap.php
```diff
@@ -171,10 +171,6 @@ public function bootCms() {
     call_user_func([$this, $func],
       $cms['path'], $this->options['user'], $this->options['httpHost']);
 
-    if (PHP_SAPI === "cli") {
-      error_reporting(1);
-    }
-
```
totten, line 177: This is more likely to be doing something. Different CMS's are likely to switch error-reporting in different ways. (*And if they have primary responsibility for processing the request, then power to them.*) But from cv-cli pov, we just want access to their APIs -- and then the CLI UX should be consistent wrt to error-handling.

Looking at this again, I'm a little on the fence. Perhaps this value should be tuned-in more to the verbosity level...

src/Application.php
```diff
@@ -43,9 +43,7 @@ public static function main($binDir) {
     $application = new Application('cv', static::version() ?? 'UNKNOWN');
 
     $application->setAutoExit(FALSE);
-    ErrorHandler::pushHandler();
     $result = $application->run();
-    ErrorHandler::popHandler();
```
totten, line 48: I believe that this was a [response](https://github.com/civicrm/cv/pull/146) to shifts in how "Symfony Console" approached error handling. Cv used `symfony/console` 2.x for a long time, and this wasn't needed. (Maybe 2.x had a more "batteries included" philosophy? Or just didn't have the bug) Somewhere around 3.x or 4.x, the default error-handling became... worse ("less opinionated"? "more vanilla"?)... e.g. it would `exit(0)` even if there was an error.

Of course, now we've switched to 5.x... maybe it's changed again...

lib/src/CmsBootstrap.php
demeritcowboy: On a live site, it's probably set to 1 to begin with. On a dev site, it's probably E_ALL. I'm not clear on why it can't just be left at whatever the surrounding environment has. But yes it could work to set error_reporting(1) here, but if you specify enough 'v's then skip setting it (and/or maybe force E_ALL with 3 v's?)

src/Application.php
demeritcowboy: Ok. I have had silence before this one existed, so it's probably not the biggest one, although there must be some error handler somewhere because normally `@` doesn't silence fatals. But removing the `@` is maybe the best bang for buck.

lib/src/Bootstrap.php
demeritcowboy: #212

PR #212, opened 2024-09-06
Trimmed down version of #211 - just remove the `@` suppression

demeritcowboy:
This can lead to silent fatals where cv just appears to do nothing, and doesn't seem needed.

One way to reproduce is:
1. Install drupal 10+civi (with php 8.1+ since you can't with anything lower anyway).
2. Using php 7.4, run something like cv vars:show. It just exits. Running with -vvv doesn't tell you the actual problem.
3. If you remove the `@`, then it tells you there's a symfony lib that isn't compatible with php 7. Also there's a message about composer.


totten:
Sounds good to me.

PR #216, opened 2024-09-20
SqlCliCommand - Tone-down warnings

totten:
We're going to start using mysql wrapper script in bknix. The warning won't apply in this case (because the wrapper script has better pass-through logic).

### DIFF
--- a/src/Command/SqlCliCommand.php
+++ b/src/Command/SqlCliCommand.php
@@ -54,7 +54,7 @@ The ENV expressions are prefixed to indicate their escaping rule:
     $datasource->loadFromCiviDSN($this->pickDsn($input->getOption('target')));
 
     $mysql = Process::findCommand('mysql');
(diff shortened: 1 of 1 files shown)


totten:
[Test failure](https://test.civicrm.org/job/CiviCRM-Cv-Test/563/console) involving `wp-demo`,`5.63` is unrelated.

I believe the problem is this: `wp-demo` pulls in latest `contactlayout` via `git`, but `contactlayout` now [requires 5.74](https://github.com/civicrm/org.civicrm.contactlayout/pull/147/files#diff-bc09ad8bd61ab113d375bf52a52eff91f7f6ccbbd7b48a0696e0e3c676916735R21). If I do a similar test-run on `wp-clean`, then it's [fine](https://test.civicrm.org/job/CiviCRM-Cv-Test/564/console).

 Probably best thing for `cv.git` is to switch testing to `wp-clean`.

PR #224, opened 2024-10-09
ErrorHandler - Present more error info about MySQL connections

totten:
Overview
--------

Present more error info about MySQL connections. This may be most useful if you're debugging a failure
in MySQL-SSL.

Technical Details
-----------------

This is a work-around for a quirk in `DB_mysqli::connect()` -- it suppresses warnings
generated by `mysqli_real_connect()`. But those can be useful!

Th work-around depends on your boot mode (Civi-first vs CMS-first).
It applies to Civi-first. For CMS-first, the CMS error-handler is active
at the critical moment -- and it doesn't have such a work-around.

I'm on the fence about merging this. This should be safe enough (very narrow
impact). But it would be cleaner (and probably better) to patch
`DB_mysqli::connect()` to simply allow warnings. (Or, failing that, buffer
the warnings -- and let them through if we hit any subsequent problems.)
Addressing that root cause would make it less sensitive to boot-mode.

Of course, warnings have been suppressed for so long... it's hard to know if
we might create some backlash from it...

### DIFF
--- a/lib/src/ErrorHandler.php
+++ b/lib/src/ErrorHandler.php
@@ -41,7 +41,14 @@ class ErrorHandler {
   }
 
   public static function onError($errorLevel, $message, $filename, $line) {
-    if ($errorLevel & error_reporting()) {
+    $errorReporting = error_reporting();
+    if (preg_match(';(/DB/|\\\DB\\\);', $filename) && static::isInCallstack('mysqli_real_connect', debug_backtrace())) {
+      // If a warning arises during the execution of `DB_mysqli::connect()` (esp during `@mysqli_real_connect()`),
+      // then the global error_reporting() is munged. Let's get our messages back...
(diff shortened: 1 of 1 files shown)


demeritcowboy:
I'd lean towards removing the `@`. I'm always leaning towards that.

But also is this a php 7 vs 8 thing? See red box at https://www.php.net/manual/en/language.operators.errorcontrol.php

totten:
> I'd lean towards removing the `@`. I'm always leaning towards that.

It is evil...

> But also is this a php 7 vs 8 thing? See red box at https://www.php.net/manual/en/language.operators.errorcontrol.php

Oh, good link. I think that means the problem is generally worse on php7. But for MySQL SSL errors (specifically),  it reports the messages with `E_WARNING`. So the `@` operator suppresses them on both php7+php8.

totten:
Close in favor of https://github.com/civicrm/civicrm-core/pull/31911

PR #226, opened 2024-11-03
Implement an encoder for generating `cv api4` args

totten:


totten:
Aside: Here's a JS port: https://gist.github.com/totten/bc853f4005e7040eccf011e317db40d8

Issue #210, opened 2024-08-30
Do not trigger PHP Deprecation warnings when booting in WordPress


michaelmcandrew:
```
admin@civicrm:/var/www/html$ cv ev 'echo 1'
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:120
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:120
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:127
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:134
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:134
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:141
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:141
1
```

This happens because wordpress does this in var/www/html/wp-includes/vars.php at line 120

```php
$is_apache = (str_contains($_SERVER['SERVER_SOFTWARE'], 'Apache') || str_contains($_SERVER['SERVER_SOFTWARE'], 'LiteSpeed'));
```

According to https://www.php.net/manual/en/reserved.variables.server.php

> When running PHP on the [command line](https://www.php.net/manual/en/features.commandline.php) most of these entries will not be available or have any meaning.

The wp cli doesn't complain becuase earlier in `wp-includes/load.php` WordPress has done this:
```
function wp_fix_server_vars() {
	global $PHP_SELF;

	$default_server_values = array(
		'SERVER_SOFTWARE' => '',
		'REQUEST_URI'     => '',
	);

	$_SERVER = array_merge( $default_server_values, $_SERVER );
```

Is there some way that we can detect is cv is booting WP and if so do this

```
	$default_server_values = array(
		'SERVER_SOFTWARE' => '',
		'REQUEST_URI'     => '',
	);

	$_SERVER = array_merge( $default_server_values, $_SERVER );
```

[...]

PR #231, opened 2025-01-14
Set $_SERVER['SERVER_SOFTWARE'] to an empty string rather than NULL

michaelmcandrew:
When using cv in WordPress, we receive deprecation warnings along the lines of

```
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:120
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:120
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:127
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:134
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:134
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:141
[PHP Deprecation] str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated at /var/www/html/wp-includes/vars.php:141
```

This is because we have set `$_SERVER['SERVER_SOFTWARE']` to `NULL` and WordPress expects it to be a string.

Given what the PHP manual say about the $_SERVER value here: https://www.php.net/manual/en/reserved.variables.server.php, I think this is a reasonable assumption. 

```
$_SERVER is an array containing information such as headers, paths, and script locations. The entries in this array are created by the web server, therefore there is no guarantee that every web server will provide any of these; servers may omit some, or provide others not listed here [...] When running PHP on the command most of these entries will not be available or have any meaning.

[...]

'SERVER_SOFTWARE'
Server identification string, given in the headers when responding to requests.
```

I'm not sure why we are setting it to NULL specifically but I am guessing it is because some code expects it to be set and that NULL seemed like a reasonable value to set it to. And Drupal does something similar here https://github.com/drupal/drupal/blob/7.x/includes/bootstrap.inc#L643-L651 so maybe that was the inspiration?

This PR is based on the assumption that setting it to an empty string is more 'correct' and that any downstream code is not relying on it being NULL (which I think is a reasonable assumption.

It supercedes https://github.com/civicrm/cv/issues/210

### DIFF
--- a/lib/src/Bootstrap.php
+++ b/lib/src/Bootstrap.php
@@ -196,7 +196,7 @@ class Bootstrap {
 
         $_SERVER['SCRIPT_FILENAME'] = $cmsBasePath . '/index.php';
         $_SERVER['REMOTE_ADDR'] = "127.0.0.1";
-        $_SERVER['SERVER_SOFTWARE'] = '';
+        $_SERVER['SERVER_SOFTWARE'] = ($cmsType === 'drupal') ? NULL : '';
         $_SERVER['REQUEST_METHOD'] = 'GET';
         if (!empty($options['httpHost'])) {
           // Hint for D7 multisite
--- a/lib/src/Bootstrap.php
+++ b/lib/src/Bootstrap.php
@@ -196,7 +196,7 @@ class Bootstrap {
 
         $_SERVER['SCRIPT_FILENAME'] = $cmsBasePath . '/index.php';
         $_SERVER['REMOTE_ADDR'] = "127.0.0.1";
-        $_SERVER['SERVER_SOFTWARE'] = NULL;
+        $_SERVER['SERVER_SOFTWARE'] = '';
         $_SERVER['REQUEST_METHOD'] = 'GET';
         if (!empty($options['httpHost'])) {
           // Hint for D7 multisite
--- a/lib/src/CmsBootstrap.php
+++ b/lib/src/CmsBootstrap.php
@@ -540,7 +540,7 @@ class CmsBootstrap {
   protected function simulateWebEnv($host, $scriptFile) {
     $_SERVER['SCRIPT_FILENAME'] = $scriptFile;
     $_SERVER['REMOTE_ADDR'] = "127.0.0.1";
-    $_SERVER['SERVER_SOFTWARE'] = NULL;
+    $_SERVER['SERVER_SOFTWARE'] = '';
     $_SERVER['REQUEST_METHOD'] = 'GET';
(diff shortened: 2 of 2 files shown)


totten:
Looks like a good update @michaelmcandrew.

I'll look into why the tests are unhappy.

totten:
civibot, test this please

totten:
OK, so I think some of the noise in the test-results is fixed. Comparing [last run](https://test.civicrm.org/duderino/batch/Cv-PR/20873748) with [recent baseline](https://test.civicrm.org/duderino/batch/Cv-PR/20873749), it does look like D7 is pretty unhappy with the `''`.

> I'm not sure why we are setting it to NULL specifically but I am guessing it is because some code expects it to be set and that NULL seemed like a reasonable value to set it to. And Drupal does something similar here https://github.com/drupal/drupal/blob/7.x/includes/bootstrap.inc#L643-L651 so maybe that was the inspiration?

Yeah, I'm pretty sure the idea came from Drupal-land (that file or some similar one, like drush).

> This PR is based on the assumption that setting it to an empty string is more 'correct' and that any downstream code is not relying on it being NULL (which I think is a reasonable assumption.

Yeah, I'm not sure either. (Obviously there's no `Apache/1.2.3` or `nginx/1.2.3` in CLI environment... one could make a case for `cv/1.2.3` -- eg `'cv/' . \Civi\Cv\Application::version()` -- as being more useful. But it's so much happenstance on what the the different systems expect). 

Guess: maybe the D7 jobs are failing on an ornery bit like this (where `NULL` and `''` evaluate differently):

https://github.com/drupal/drupal/blob/7.103/includes/bootstrap.inc#L3817-L3822

totten:
@michaelmcandrew Tests are happy now.

Let me know if this revision works for you.

(Aside: I'm pretty sure I've seen those cv-wp warnings before, so very glad there's a fix. But I'm having trouble provoking them right now. I'll take your word if it seems good.)

michaelmcandrew:
just to say that this has filtered through to our tools now - I'm enjoying the silence :)

totten:
Woot! (Hopefully, an exclamation mark doesn't break the silence too much. :wink: )

PR #233, opened 2025-02-05
Debugging - Print details about SQL errors (more compat)

totten:
Overview
-------------

Follow-up to #230, which improves support for:

* Handling exceptions when the command is executed as `cv.phar` (instead of `cv` source)
* Handling chained exceptions (`getPrevious()`)

Before
---------

If you invoke `cv` as a PHAR file, or if you have chained exceptions, then the DB errors are reductive:

Screenshot from 2025-02-04 23-52-54

After
------

Displayed with more debug info:

Screenshot from 2025-02-04 23-53-08

Comments
--------------

* `getCause()` and `getPrevious()` sound similar, but actually differ:
    * `getPrevious()` is a PHP stdlib thing (v7+). It points to the earlier `Exception` which provoked the current exception.
    * `getCause()` is a PEAR thing. It unwraps the `PEAR_Error` that provoked the `Exception`

### DIFF
--- a/scoper.inc.php
+++ b/scoper.inc.php
@@ -21,6 +21,7 @@ return [
   ],
   'exclude-classes' => [
     '/^(CRM_|HTML_|DB_|Log_)/',
+    '/^PEAR_(Error|Exception)/',
     'DB',
     'Log',
     'JFactory',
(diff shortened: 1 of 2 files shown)


totten:
Technically, either one of the commits is sufficient for fixing the PHAR compat. But they're also fine together.

Re: test failures --They indicate an unrelated outage for `civicrm.org/extdir`.

colemanw:
Nice, thanks.

PR #235, opened 2025-02-12
Miscellaneous cleanups

totten:
Each commit has its own description


totten:
All test failures are pre-existing/unrelated issue on external server.

PR #236, opened 2025-02-12
Improve handling of active URLs (esp HTTPS construction)

totten:
Overview
----------

Improve the integration between `cv`, its URL inputs, and the common `$_SERVER` variables. Specifically, if you give the `--url` option, then it will set variables like `HTTP_HOST`, `HTTPS`, and `SERVER_PORT`

```
$ cv ev --url=https://example.com:4444 'echo $_SERVER["HTTP_HOST"];'
example.com:4444

$ cv ev --url=https://example.com:4444 'echo $_SERVER["SERVER_NAME"];'
example.com

$ cv ev --url=https://example.com:4444 'echo $_SERVER["SERVER_PORT"];'
4444

$ cv ev --url=https://example.com:4444 'echo $_SERVER["HTTPS"];'
on
```

By setting these server-variables, we should give better hints to any URL-generating functions in CMS or Civi.

The older options `--cms-base-url` and `--hostname` are retained as unpublicized aliases. The new `--url` option is the combined+expanded replacement.

Example Problem
------------------

In https://github.com/civicrm/civicrm-core/pull/31271, @seamuslee001 mentions the use-case where one installs a WordPress site with an HTTPS address:

```bash
cv core:install --cms-base-url='https://example.com/'
```

This URL is properly recorded in `CIVICRM_UF_BASEURL`. However, when calling APIs to generate URLs (such as WP's `plugin_dir_url()`), those functions may sometimes compose `http://` URLs _even though_ the user indicated that `https://` is valid. (And if `https://` is valid, then it really should be used.)

For `plugin_dir_url()`, the reason is that WP is inspecting the active request (`$_SERVER` and its properties, `HTTP_HOST`, `HTTPS`, `SERVER_PORT`, etc); these don't include any HTTPS flags, so we get basic HTTP.

It seems that we should provide stronger direction on how these server-vars behave in CLI context.

Before
-------

* All `cv` commands accept input from `--hostname example.com` and env-var `HTTP_HOST=example.com`. 
    * These do feed into `$_SERVER`, but they don't indicate whether HTTPS is supported.
* In `cv core:install`, there is also an option `--cms-base-url` which accepts a fully formed URL. 
    * This conveys more information (including HTTPS), but it doesn't feed into `$_SERVER`.
* If you do not specify a URL via `--hostname` or `HTTP_HOST`, then `$_SERVER` is devoid of information about HTTP/SSL. Consequently, you are liable to emit `http://` URLs and `localhost` URLs.

After
-----

* The old options `--hostname` and `--cms-base-url` are deprecated. They function as aliases for `--url`.
* All `cv` commands accept input from  `--url` (`-l`) and env-vars `HTTP_HOST`, `HTTPS`, `SERVER_PORT`.
    * This feeds into `$_SERVER`, and it also indicates HTTPS support.
    * As before, you can give a bare hostname -- but you can also give a fully formed URL.

        ```bash
        cv ev 'echo $_SERVER["HTTP_HOST"];' --url=example.com
        cv ev 'echo $_SERVER["HTTP_HOST"];' --url=https://example.com
        cv ev 'echo $_SERVER["HTTP_HOST"];' --url=http://example.com
        ```

        If the given URL is `https://`, then you can expect most URL-composition to enable HTTPS.

        This works for `core:install` (CMS-first boot) and all regular commands (Civi-first or CMS-first boot).

    * The new option is shorter and easier to type. It also gives a nod to folks who use wp-cli or drush:
        | Cv Options | Similar To | Comments |
        | -- |-- |-- |
        | `cv --url=XXX` | `wp --url=XXX` | In vernacular speech, "URL" is common. |
        | `cv --uri=XXX` | `drush --uri=XXX`| Unpublicized alias. "URI" is less vernacular. |
        | `cv -l XXX`  | `drush -l XXX` | |

* If you do not specify `--url` (or similar env-vars), then the behavior varies:
    * If Civi boots first, then it populates `$_SERVER` using `CIVICRM_UF_BASEURL`. (*You will get HTTPS if the BASEURL has HTTPS.*)
    * If CMS boots first, then it populates `$_SERVER` using `http://localhost`. (This doesn't seem ideal -- it will still tend toward HTTP. But it is the status-quo, and it's trickier to find a fix.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,14 +228,14 @@ Bootstrap
 > ___NOTE___: In absence of a configuration variable, the __Automatic__ mode will behave like `CIVICRM_SETTINGS="Auto"` (in v0.3.x).
   This is tentatively planned to change in v0.4.x, where it will behave like `CIVICRM_BOOT="Auto://."`
 
-Additionally, some deployments handle multiple sites ("multisite"/"multidomain"). You should target a specific site using `--hostname` or `HTTP_HOST`.
+Additionally, some deployments handle multiple sites ("multisite"/"multidomain"). You should target a specific site using `--url` or `HTTP_HOST`.
 
 Here are a few examples of putting these together:
 
 ```bash
-## Use --hostname for a domain
+## Use --url for a domain
 export CIVICRM_BOOT="WordPress:/$HOME/public_html/"
-cv --hostname='www.example.org' ext:list -L
+cv --url='https://www.example.org' ext:list -L
 ```
 
 ```bash
@@ -246,9 +246,9 @@ cv ext:list -L
 ```
 
 ```bash
-## Use --hostname for a subfolder
+## Use --url for a subfolder
 export CIVICRM_BOOT="WordPress:/$HOME/public_html/"
-cv --hostname='www.example.org/nyc' ext:list -L
+cv --url='www.example.org/nyc' ext:list -L
 ```
 
 Autocomplete
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -113,7 +113,7 @@ The `\Civi\Cv\Cv` facade provides some helpers for implementing functionality:
 You can register new subcommands within `cv`. `cv` includes the base-class from Symfony Console, and its adds another base-class. Compare:
 
 * `CvDeps\Symfony\Component\Console\Command\Command` is the original building-block from Symfony Console. It can define and parse CLI arguments, but it does *not* bootstrap CiviCRM or CMS. It may be suitable for some basic commands. Documentation is provided by upstream.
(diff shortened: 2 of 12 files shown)


totten:
Rebased.

I've done some light testing on Drupal multisite.  All seemed good:

* It successfully installed new subdomains. 
* And if you called with `--url=subdomain.example.com`, then it loaded the corresponding `civicrm.settings.php`.
* As expected, `--hostname` still works as well (just with a deprecation notice).
* When using the [`@site` aliases](https://gist.github.com/totten/8241b80440221555c0051d4f3447fa40), it still worked.
    * If the alias specifies the deprecated `--hostname` option, then it works (with a warning).
    * If the alias specifies the new `--url` option (or the `HTTP_HOST` env-var), then it works.

Test failures are unrelated/external.

PR #240, opened 2025-02-20
Plugins - Add 'basic-alias' and 'alias-cmd' to cv.phar

totten:
Overview
----------

Site aliases allow you run a `cv` subcommand on a particular site without manually navigating through the filesystem. Compare:

```bash
## Explicitly navigate the filesystem
cd /home/myuser/projects/wombat.org/web && cv api4 Contact.get

## Explicitly use SSH, navigate the filesystem and set various boot options
ssh server.wombat.org
cd /home/myuser/projects/wombat.org/web
cv api4 Contact.get \
  --url='https://www.wombat.org/' \
  --level=cms-first \
  --user=admin

## Use an alias
cv @wombat api4 Contact.get
```

In the last example, it uses an alias (`@wombat`). The alias can define all these options (path, URL, etc) so that you don't need to type them into every command.

This functionality was previously demonstrated by implementing  the [`basic-alias` plugin](https://gist.github.com/totten/8241b80440221555c0051d4f3447fa40) that was published via gist, but this is fairly obscure. This PR adds `basic-alias` plugin as a built-in feature, and it also introduces a complementary plugin (`alias-cmd`) for managing aliases via CLI.

Before
-------

* You can manually define bootstrap options like `cd`, `--cwd`, `--user`, `--url`.

* You can download the alias plugin and create alias-files by hand.

After
-----

There is a built-in alias-plugin and helper commands.

To define a new alias, run the `alias:add` command:

```bash
## Start an interactive questionnaire
$ cv alias:add

## As above, but pre-fill some of the answers
$ cv alias:add wpmaster /var/www/wpmaster/web
```

This will ask questions about remote (SSH), multisite (base URL), default user-name, etc. It will ultimately create a file (such as `$HOME/.cv/alias/wpmaster.json`).

Similarly, you can browse available aliases:

```bash
cv alias:list
```
```
+----------+-------+---------------------------------------+
| name     | type  | config                                |
+----------+-------+---------------------------------------+
| dmaster  | basic | /Users/totten/.cv/alias/dmaster.json  |
| wpmaster | basic | /Users/totten/.cv/alias/wpmaster.json |
+----------+-------+---------------------------------------+
```

Finally, you can run commands with the alias, as in:

```bash
cv @wpmaster status
cv @wpmaster ext:list -Li
```

Comments
-----------

There's a sample build at https://think.hm/tmp/cv-aliases-2025-02-26-0.phar.

The general reason why `basic-alias` is a plugin is that it would be handy to make new variations on the functionality. There's a lot of subjectivity in what features should be provided for aliases. For example:

* If you use `civibuild`, then you might want to automatically recognize aliases for all `BUILDKIT/build/*.sh` files.
* If you want contextual aliases within a build (`@staging`, `@prod`), then you need to seek-out a context-sensitive config-file.

In any case, the absence of a reference-implementation makes the whole thing seem esoteric.

With this approach, we can kind of have it both ways:

* There is a default/out-of-the-box/concrete option for site-aliases.
* The default option is structured as a plugin -- so the wiring is plugin-friendly, and it can be adapted to other techniques.

### DIFF
--- a/box.json
+++ b/box.json
@@ -2,6 +2,7 @@
     "chmod": "0755",
     "directories": [
         "lib/src",
+        "lib/plugin",
         "src"
     ],
     "finder": [
--- a/lib/plugin/alias-cmd.php
+++ b/lib/plugin/alias-cmd.php
@@ -58,7 +58,7 @@ class AliasAddCommand extends CvCommand {
     $this
       ->setName('alias:add')
       ->addArgument('name', InputArgument::OPTIONAL, 'Alias name')
-      ->addArgument('path', InputArgument::OPTIONAL, 'Local path to the instance (web-root)')
+      ->addArgument('local-path', InputArgument::OPTIONAL, 'Local path to the instance (web-root)')
       ->setDescription('Interactively create a new @alias')
       ->setBootOptions(['auto' => FALSE]);
   }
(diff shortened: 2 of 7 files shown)


totten:
Well, got a +1 on chat, and yolo

Issue #195, opened 2024-06-02
Can't run DB upgrades - permission error


kenahoo:
I originally asked this as a question at Stack Overflow: https://civicrm.stackexchange.com/questions/48137/upgrade-5-66-0-5-66-2-fails-with-permission-error .

I'm using CiviCRM with Wordpress 6.5.3 and MySQL 5.7.44.

I'm attempting to upgrade my CiviCRM from 5.66.0 to 5.66.2.  When I get to the DB upgrade stage, when I hit `/wp-admin/admin.php?page=CiviCRM&q=civicrm/upgrade&reset=1` -> `Upgrade Now` -> `OK`, I just see an empty progress bar and the message `[Executed: undefined]`.

If I switch to the command-line `cv` (version `v0.3.48`), and turn on maximum debugging output, I see this:

```
singersin@s483:~/www/wp$ cv upgrade:db -vvv

Box Requirements Checker
========================

> Using PHP 8.1.28
> PHP is using the following php.ini file:
  /home/singersin/www/wp/php.ini

> Checking Box requirements:
  ✔ The application requires a version matching ">=7.3.0".
  ✔ The application requires the extension "json".
  ✔ The package "psy/psysh" requires the extension "json".
  ✔ The package "nikic/php-parser" requires the extension "tokenizer".
  ✔ The package "psy/psysh" requires the extension "tokenizer".
  
                                                                                                                                                                                                                      
 [OK] Your system is ready to run the application.                                                                                                                                                                    
                                                                                                                                                                                                                      

[BootTrait:debug] Start
[BootTrait:debug] Attempting to set verbose error reporting
[BootTrait:debug] Call standard cv bootstrap
[Bootstrap:debug] Options: {
    "dynamicSettingsFile": "phar:///home/singersin/private/bin/cv/lib/settings_location.php",
    "env": "CIVICRM_SETTINGS",
    "prefetch": true,
    "settingsFile": null,
    "search": true,
    "cmsType": null,
    "httpHost": "",
    "output": {}
}
[Bootstrap:debug] Find settings file
[Bootstrap:debug] Load supplemental configuration for "/home/singersin/www/wp/wp-content/uploads/civicrm/civicrm.settings.php"
[Bootstrap:notice] Find CMS root for "/home/singersin/www/wp"
[Bootstrap:notice] Found "wp" in "/home/singersin/www/wp"
[Bootstrap:notice] Simulate web environment in CLI
[...]

PR #243, opened 2025-03-17
Use XDG_STATE_HOME environment variable as a preference above HOME

jaapjansma:
Fix for #195 

One can set the environment variable `XDG_STATE_HOME` 
This will take precedence over the `HOME` environment variable.

### DIFF
--- a/lib/plugin/basic-alias.php
+++ b/lib/plugin/basic-alias.php
@@ -89,6 +89,9 @@ class AliasFinder {
     elseif (getenv('USERPROFILE')) {
       array_unshift($dirs, getenv('USERPROFILE') . '/.cv/alias');
     }
+    if (getenv('XDG_CONFIG_HOME')) {
+      array_unshift($dirs, getenv('XDG_CONFIG_HOME') . '/cv/alias');
+    }
     return $dirs;
   }
 
--- a/lib/src/Config.php
+++ b/lib/src/Config.php
@@ -45,6 +45,9 @@ class Config {
     if (getenv('CV_CONFIG')) {
       return getenv('CV_CONFIG');
     }
+    elseif (getenv('XDG_STATE_HOME')) {
+      return getenv('XDG_STATE_HOME') . '/.cv.json';
+    }
     elseif (getenv('HOME')) {
       return getenv('HOME') . '/.cv.json';
     }
--- a/lib/src/Config.php
+++ b/lib/src/Config.php
@@ -43,17 +43,33 @@ class Config {
    */
   public static function getFileName() {
     if (getenv('CV_CONFIG')) {
+      // The user has specifically told us where to go.
       return getenv('CV_CONFIG');
     }
-    elseif (getenv('XDG_CONFIG_HOME')) {
-      return getenv('XDG_CONFIG_HOME') . '/.cv.json';
+
+    // We have to figure out where to go. There are a couple plausible locations...
+    $candidates = [];
+    if (getenv('XDG_CONFIG_HOME')) {
+      $candidates[] = getenv('XDG_CONFIG_HOME') . '/.cv.json';
     }
-    elseif (getenv('HOME')) {
-      return getenv('HOME') . '/.cv.json';
+    if (getenv('HOME')) {
+      $candidates[] = getenv('HOME') . '/.cv.json';
     }
-    else {
-      throw new \RuntimeException("Failed to determine file path for 'cv.json'.");
+
+    // Prefer the first extant config file...
+    foreach ($candidates as $candidate) {
+      if (file_exists($candidate)) {
+        return $candidate;
+      }
     }
+
+    // Or if there is no extant file, then use the first plausible suggestion...
+    if (isset($candidates[0])) {
+      return $candidates[0];
+    }
+
+    throw new \RuntimeException("Failed to determine file path for 'cv.json'.");
+
   }
 
 }
--- a/lib/src/Config.php
+++ b/lib/src/Config.php
@@ -45,8 +45,8 @@ class Config {
     if (getenv('CV_CONFIG')) {
       return getenv('CV_CONFIG');
     }
-    elseif (getenv('XDG_STATE_HOME')) {
-      return getenv('XDG_STATE_HOME') . '/.cv.json';
+    elseif (getenv('XDG_CONFIG_HOME')) {
+      return getenv('XDG_CONFIG_HOME') . '/.cv.json';
     }
     elseif (getenv('HOME')) {
       return getenv('HOME') . '/.cv.json';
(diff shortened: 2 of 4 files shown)


totten:
Thanks for moving this forward, @jaapjansma!

The test-failures seem to be some unrelated issue in `master`. (I haven't yet figured out a sufficient routine to focus-in on the problem - but will look at it.)

I wonder if you've had a chance to read the definitions from  https://specifications.freedesktop.org/basedir-spec/latest/#variables -- there are a few variables, and the choice of vars somewhat subjective (but not completely). 

Like here's my impression

* `UpgradeDbCommand::getUpgradeFile()` - Sounds like closest fit to `XDG_STATE_HOME`?
* `Config::getFileName()` and `CvPlugins` - These sound like `XDG_CONFIG_HOME` might be closer fit?

jaapjansma:
I agree with you for following the standard. However using `cv` on a well used hosting platform as CiviHosting one needs to set two environment variables to get `cv` working. Because on CiviHosting those environment variables are not  set. 

So this requires additional documentation and makes the tool probably less user friendly.

jaapjansma:
I have updated the PR to reflect the standard of XDG Base Directories

totten:
Great. Thanks for moving this forward @jaapjansma.

I'm adding a revision (*and plan to merge assuming tests pass*) to try to solve a little puzzle:

* XDG vars should take priority - but historically, they haven't been factor.
* Some people already have `~/.cv.json`
* It's hard to say when, exactly, the XDG vars are preset. (Ex: In my current Ubuntu version, the distro presents ~8 XDG vars -- but not specifically `XDG_CONFIG_HOME`.)
* If one already has `~/.cv.json`, and if one already has XDG vars, then the revision could block loading of the (previously-valid) `~/.cv.json`.

So with the new commit, it will check a prioritized list of candidates (`XDG_CONFIG_HOME/.cv.json` then `HOME/.cv.json`) and use an existence-check. This should mean that:

* If you already have `.cv.json` (in either location), then it will prefer to use the existing file (wherever that is).
* If you don't already have `.cv.json`, it will use the first plausible location (either `XDG_CONFIG_HOME` or `HOME`).

(*For the upgrade-log case --  the data is temporary/transactional, so it doesn't matter if the path switches...*)

demeritcowboy:
It seems this just broke CiviCarrot. My initial guess is that `cv vars:fill` now creates/fills something other than `~/.cv.json`, but carrot's assumption was that it would be `~/.cv.json`. i.e. in github actions XDG_CONFIG_HOME may not be `~`

Assuming this is true, I don't know how many other environments assume `~` is the same as XDG_CONFIG_HOME.

demeritcowboy:
This is an easy update for carrot, and it's running now. But as noted I don't know how many other environments's scripts are unaware the XDG stuff exists (I wasn't), and just use `~`.

totten:
@demeritcowboy Ah, thanks for reporting back.

> carrot's assumption was that it would be `~/.cv.json`

So the implication is that you automatically spin-up a new system and have some other script-y-bits which manipulate the `.cv.json`? Along the lines of (pseudocode)...

```bash
function on_create_new_system() {
  ...
  git clone https://github.com/civicrm/civicrm-core
  ...
  cv core:install ...
  ...
  cv vars:fill
  jq '.sites["/my/site1"].SOME_FIELD = "newvalue"'  ~/.cv.json | sponge ~/.cv.json
  ...
}
```

If I understand -- #243 should be OK if you manually use `vars:fill` and manually edit the displayed JSON file -- the problem is if `vars:fill` is part of a script (*which isn't something I ever considered doing - but I can see how it fits for carrot*). Possible reactions are like:

* (1) Update script -- Pre-configure `CV_CONFIG=$HOME/.cv.json`, or...
* (2) Update script -- Pre-create `~/.cv.json` as an empty placeholder, or...
* (3) Update script -- Parse the output of `cv vars:fill` to determine the real filename, or...
* (4) Update script -- Pass the data to `cv vars:fill` (ex: `echo '{"FOO":"bar"}' | cv vars:fill --file=/dev/stdin`) so that you don't need to touch the file directly.
    * (4b) (Or if "fill" behavior isn't quite the right fit, then we could update both `cv` and the script -- add a subcommand like `cv vars:set FOO=bar`.)
* (5) Partial-revert -- keep the parts which use XDG_STATE_HOME for upgrade-logs (needed for #195) but not the part for XDG_CONFIG_HOME (`.cv.json`).

Unfortunately, it's hard to tell if there other scripts that call `vars:fill`. FWIW, in `universe`, I don't see any other scripted references to `vars:fill` or `cv.json`. (The search covers a handful of docker projects, and there area couple dozen extensions with their own `.github/` or `.gitlab/`.)

Maybe we hold as-is for the moment -- but if someone comes along to comment on a second script that's impacted, then we can consider one of the other interventions?

demeritcowboy:
Carrot basically follows the dev docs setup steps: https://docs.civicrm.org/dev/en/latest/testing/#setup

```
$ cv vars:fill
$ vi ~/.cv.json
```

just obviously doesn't use vi.

So where vars:fill creates the file has potentially changed, and if you're unaware the XDG vars are set on your system or even exist then that description is no longer correct.

jaapjansma:
Maybe we can show the path upon `cv vars:show` command, is that an idea and then update the docs?

totten:
Yeah, I think adding some more commands like `cv vars:set` or `cv vars:show` would make sense for that kind of semi-automated configuration -- so that you can script it without needing to know the specific file-format. (Similar how to `composer` allows you to do scripted configuration via `composer config` -- you usually don't need scripts to edit `composer.json`.)

Stepping back for a second -- we also now have site-aliases (#240). In a strict technical sense, `.cv.json` and site-aliases are completely separate things which...

* Track different variables (`_CV[DEMO_PASS]` vs `SERVER[HTTP_HOST]`)
* Load at different moments, in different layers
* Read from different sources (`~/.cv.json` vs `~/.cv/alias/*.json`) with different keys (`$config['sites']['/path/to/civicrm.settings.php']` vs `@aliasname`)
* And basically address different problems (*phpunit vs sysadmin*).
 
And yet... conceptually... they look __really, really__ similar (*stored config-vars keyed by a site-identifier*) and have some similar workflow needs (*needing helpers for  interactive-setup or scripted-setup*). I suspect there will be some way to reconcile them into the same mechanism... but it's probably a multiple-step process (not planned or scheduled right now).

My point is that `~/.cv.json` may change more in the future. There are ways to deal with that:

* Be lazy. Don't do anything more to `.cv.json` until there's a clearer picture on whether/how to unify with aliases.
* Be active. Add some more commands (`cv vars:set`) to abstract away from the format. (When changes come, we can edit commands to support new formats.)

jaapjansma:
I did a follow up PR showing the `cv.json` file location when running `cv vars:show`. See #248

PR #248, opened 2025-04-10
Show the config file location of cv.json with the command vars:show 

jaapjansma:
This PR is a follow up on https://github.com/civicrm/cv/pull/243#issuecomment-2782715117

It will show the location of `cv.json` file when running `cv vars:show`

See screenshot below


totten:
Thanks @jaapjansma

PR #252, opened 2025-05-23
Implement Joomla CMS boot routine

seamuslee001:
This implements cmsBoot function for Joomla and associated matters. 

I have tested this on a J4 site and it seems to work. It doesn't work for a J3 site but that is EOL so maybe ok?, I haven't tested what it does on a J5 site yet.

one element I wasn't sure about is if in L230 I should use $this->bootedCms['path'] or JPATH_BASE. The way it is coded up now it loads the administrator civicrm.settings.php whereas using $this->bootedCms['path'] would load up the front end (site) civicrm.settings.php

### DIFF
--- a/lib/src/CmsBootstrap.php
+++ b/lib/src/CmsBootstrap.php
@@ -225,7 +225,24 @@ class CmsBootstrap {
       \Civi\Cv\Bootstrap::singleton()->boot();
       $this->loginStandaloneUser();
     }
-    // else if Joomla weirdness, do that
+    elseif ($this->bootedCms['type'] === 'Joomla') {
+      if (!defined('CIVICRM_SETTINGS_PATH')) {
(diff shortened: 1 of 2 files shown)


totten:
Sweet! I've been playing around with this on a local J4, and it appears to work. 🎉 

> one element I wasn't sure about is if in L230 I should use $this->bootedCms['path'] or JPATH_BASE. The way it is coded up now it loads the administrator civicrm.settings.php whereas using $this->bootedCms['path'] would load up the front end (site) civicrm.settings.php

Yeah, that's a good point. I think you've made the right call as-is. Either choice runs the risk of being "right for some use-cases, wrong for other use-cases". But the important difference is that `CmsBootstrap.php` aims to replace/interoperate with `Bootstrap.php` -- and *that* also uses the administrator `civicrm.settings.php`.

So this approach seems more consistent.

Issue #246, opened 2025-04-09
add cv cron command


michaelmcandrew:
Following discussion here: https://github.com/civicrm/civicrm-docker/issues/28#issuecomment-2788674265

What do you think about adding a `cv cron` command that is a wrapper around `cv api3 Job.execute` and also picks up a new setting in CiviCRM called CRON_USER which is the user that is used to execute the command.

This would simplify sysadmin because we would be able to call cron for sites without knowing anything about the users installed on those sites. We currently need to maintain a mapping of sites and their cron users outside of CiviCRM in order to trigger cron.

Aside: should we really need to pass a cms user to execute a cron job in 2025?

michaelmcandrew:
@mlutfy - any thoughts on this given your experience with Spark?

mlutfy:
@michaelmcandrew Any simplification is welcome, for sure.

It would be great to have the equivalent of a fake user for cron, which cannot be used for logins, and has full permissions. And where Scheduled Reminders or other scheduled job activities are not all "Admin, Admin".

Spark uses Aegir, and we have a lot of monitoring around cron, because admins change the username, delete the account, etc. (Spark locks the superadmin  account down, but my other verticals do not)

Aegir stores the cron username in the site settings, part of the Ansible inventory. Ansible is then used to generate systemd timers/service units.

totten:
> should we really need to pass a cms user to execute a cron job in 2025?

Yeah, I also feel like it's poor-UX/AX for crons to require a user. I'm not really clear on how much trouble there would be to omit them.

FWIW, if it's just a matter of ensuring that `CRM_Core_Permission::check('foo')` will pass... then we could try [something like this](https://github.com/civicrm/civicrm-core/blob/1dd3d9e2f6de1189cf4211173e8b940e2dafa5c7/ext/afform/mock/tests/phpunit/api/v4/Afform/AfformTestCase.php#L27-L28) before calling `Job.execute`:

```php
    \CRM_Core_Config::singleton()->userPermissionTemp = new \CRM_Core_Permission_Temp();
    \CRM_Core_Config::singleton()->userPermissionTemp->grant('administer CiviCRM');
```

Other theories (*speculative*) for how user might be important to cronjobs:

* _Logs_: If a new record is created/modified by a cron, how is it attributed in the logs?
    * We could protentially autocreate a `Contact` for this purpose and set the corresponding `userId` in session.
[...]

PR #254, opened 2025-05-25
issue #246: add core cron command

mlutfy:
Adds a `cv core` command, based on the code proposed by @totten in #246 

I tested with a job that creates an activity, and it's neat, because it looks like this:

image

I was wondering about output and errors, but since it uses Job.execute, exceptions are hidden (in the job logs, which gets monitored separately). If there was a PHP fatal, we would see it on screen.

Cron output is usually logged, so it's nice to be terse, only output if there are PHP fatal errors.

### DIFF
--- a/src/Command/CronCommand.php
+++ b/src/Command/CronCommand.php
@@ -20,8 +20,16 @@ class CronCommand extends CvCommand {
 
   protected function execute(InputInterface $input, OutputInterface $output): int {
     if (empty($input->getOption('user'))) {
-      \CRM_Core_Config::singleton()->userPermissionTemp = new \CRM_Core_Permission_Temp();
-      \CRM_Core_Config::singleton()->userPermissionTemp->grant('all CiviCRM permissions and ACLs');
+
+      // Grant access to all permissions for the current process.
+      // Ex: `Job.process_mailing` needs permission to lookup recipients.
+      \CRM_Core_Config::singleton()->userPermissionTemp = new class extends \CRM_Core_Permission_Temp {
+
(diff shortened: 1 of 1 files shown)


mlutfy:
cc @michaelmcandrew @ufundo

mlutfy:
Yep I can imagine that people will have various use-cases eventually (multi-domains, non-default org user). This provides a simpler way to get started with cron, and if we do add more options, I don't think it will break the implementation for early-adopters (so I think those options can be discussed separately).

totten:
> I tested with a job that creates an activity, and it's neat, because it looks like this:

Oooh 🤩 

> Cron output is usually logged, so it's nice to be terse, only output if there are PHP fatal errors.

Yep, agree it should be quiet by default.

For general UX/DX in CLI, it is also nice to support a verbose mode (e.g. print out the tasks that are running -- so that you can see where a failure arises). Not urgent. (But I've split off a separate PR -- https://github.com/civicrm/civicrm-core/pull/32966 -- to facilitate verbose-mode.)

------

I did some spot-checking to see if the permission trick was working... and it wasn't. (For example, `cv cron` would fail to run `Job.process_mailing`.)

I think the problem is that `CRM_Core_Permission_Temp` doesn't work with magical permisisons like `all CiviCRM permissions and ACLs`.

So I'm pushing up another formulation of `CRM_Core_Permission_Temp`, which is working better for me. (I'll try to find another `r-run` scenario to confirm.)

mlutfy:
@totten Very cool, thank you!

PR #261, opened 2025-08-14
api4 - Display columns with shifty names (eg `COUNT(*)` vs `COUNT:`)

totten:
Consider a request like this:

```
cv api4 Contact.get +s 'COUNT(*)' -T
```

Internally, when handling this request, the current versions of CiviCRM will return a column named `COUNT:`. This update makes those results displayable.

## Before

Result appears blank:

```
+----------+
| COUNT(*) |
+----------+
|          |
+----------+
```

## After

Result shows data:

```
+--------+
| COUNT: |
+--------+
| 204    |
+--------+
```

Of course, that's kinda weird too -- but it's a truer representation of the result.

## Comments

This also fixes tabular display when using `AS`:

```
$ cv api4 Contact.get '{"select":["COUNT(*) AS the_count"]}' -T
+-----------+
| the_count |
+-----------+
| 204       |
+-----------+
```

### DIFF
--- a/src/Command/Api4Command.php
+++ b/src/Command/Api4Command.php
@@ -146,12 +146,19 @@ NOTE: To change the default output format, set CV_OUTPUT.
 
     $out = $input->getOption('out');
     if (!in_array($out, Encoder::getFormats()) && in_array($out, Encoder::getTabularFormats())) {
-      if (!empty($params['select'])) {
-        $columns = $params['select'];
-      }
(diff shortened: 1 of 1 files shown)


colemanw:
Looks good. There's really no scenario in which Api4 *should* return a different list of columns per row (and if it did, that would kinda defeat the idea of a table display anyway).

PR #278, opened 2025-12-15
Add a password-reset command for Standalone

mlutfy:
Adds a password-reset command for CiviCRM Standalone.

Eventually the command could be used for setting a new password as well. Keeping it simple for now.

### DIFF
--- a/src/Command/PasswordResetCommand.php
+++ b/src/Command/PasswordResetCommand.php
@@ -1,0 +1,82 @@
+<?php
+namespace Civi\Cv\Command;
+
+use Civi\Cv\Encoder;
+use Civi\Cv\Util\ExtensionTrait;
+use Civi\Cv\Util\Process;
+use Civi\Cv\Util\StructuredOutputTrait;
+use Civi\Cv\Util\UrlCommandTrait;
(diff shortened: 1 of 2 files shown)


src/Command/PasswordResetCommand.php
```diff
@@ -0,0 +1,82 @@
...
+
+    $token = \Civi::service('crypto.jwt')->encode([
+      'exp' => $expires,
+      'sub' => "uid:$uid",
+      'scope' => \Civi\Api4\Action\User\PasswordReset::PASSWORD_RESET_SCOPE,
+    ]);
+    \Civi\Api4\User::update(FALSE)
+      ->addValue('password_reset_token', $token)
+      ->addWhere('id', '=', $uid)
+      ->execute();
```
ufundo: Couldn't this just be?
```suggestion
    $token = \Civi\Api4\Action\UserPasswordReset::updateToken($uid, $input->getOption('expires', 60));
```

Then if the underlying logic changes a bit it stays consistent?

Or even add a User.generatePasswordResetLink to Api4? I think ideally `cv` would stick to public apis to avoid implementation linkages.

PR #290, opened 2026-04-21
Don't assume the existence of CIVICRM_UF_DSN

MegaphoneJon:
If your `civicrm.settings.php` doesn't have a `CIVICRM_UF_DSN`, `cv vars:show` returns a fatal error.

Since ideally a standalone Civi doesn't have a `CIVICRM_UF_DSN`, `cv` should support the non-existence of it.


colemanw:
Looks good, thanks @MegaphoneJon

